PR #1, opened 2016-09-23
Bug 1305045 - Make scripts/push_apk.py usable without Mozharness

JohanLorenzo:


mozapkpublisher/get_apk.py
```diff
@@ -0,0 +1,206 @@
...
+        apk_checksum = self.file_sha512sum(apk_file)
+
+        if checksum == apk_checksum:
+            logger.info("APK checksum check succeeded!")
+            self.download_complete(apk_file, checksum_file)
+        else:
+            shutil.rmtree(self.download_dir)
+            logger.fatal("Downloading " + apk_file + " failed!")
+
+    def file_sha512sum(self, file_path):
```
escapewindow: A number of methods could be turned into standalone functions.  This is a method that never references `self`, so it would be just as valid as a function.  A full pep8 check would highlight the non-use of `self`.

In mozharness I made everything part of an object to keep everything together, but ended up with monolithic *Script objects with too many mixins.  Post-mozharness I've moved to more clean architecture functional programming.  This doesn't block, just mentioning because I noticed.

mozapkpublisher/get_apk.py
```diff
@@ -0,0 +1,206 @@
...
+
+    # Gets called once download is complete
+    def download_complete(self, apk_file, checksum_file):
+        logger.info(apk_file + " has been downloaded successfully")
+        os.remove(checksum_file)
+
+    # Called if download fails due to 404 error or some other connection failure
+    def download_error(self):
+        self.cleanup()
+        logger.fatal("Download failed!")
```
JohanLorenzo: Thanks for the context. That made me realize `download_error()` is not used anymore in the current implementation. I looked up the other locations of `fatal()` and raised errors there.

mozapkpublisher/get_apk.py
JohanLorenzo: Agreed, function moved to `utils`. For the record neither `flake8` nor `pep8` reported self as being non-used. I guess that's a matter of configuration. I'll look into that as well.

mozapkpublisher/get_apk.py
```diff
@@ -0,0 +1,206 @@
...
+        if arch == "all":
+            self.download_all(version, build, locale)
+        else:
+            self.download(version, build, arch, locale)
+
+    def run(self):
+        self.download_apk()
+
+
+if __name__ == '__main__':
```
JohanLorenzo: Alright! Will do a in follow up.

mozapkpublisher/utils.py
```diff
@@ -0,0 +1,15 @@
+import requests
```
JohanLorenzo: Understood. I believe the scope of these script won't be larger than downloading more than a couple of apks and checksums per run. Also, as the scripts used to be python 2, I think it's better to keep a transition period where both py2 and py3 are supported.

mozapkpublisher/push_apk.py
```diff
@@ -0,0 +1,161 @@
...
+
+        # Retrieve the mapping
+        self.translationMgmt.load_mapping()
+
+        # For each files, upload it
+        for apk_file in apk_files:
+            apk_file_name = apk_file.name
+            try:
+                # Upload the file
+                apk_response = service.edits().apks().upload(
```
JohanLorenzo: I agree. This part also breaks the Law of Demeter. Let's make this part cleaner in a follow, so this PR only focuses on being mozharness-less. 

Btw, nice talk. I've never watched it before, and it's very inspirational!

PR #2, opened 2016-10-26
Upgrade lower tracks of Google Play

JohanLorenzo:
Inspired from [this feature](https://github.com/Triple-T/gradle-play-publisher#untrack-conflicting-versions) [implemented there](https://github.com/Triple-T/gradle-play-publisher/blob/8a2fa9064bc50fcd276b3c3d00381d2f4f873296/src/main/groovy/de/triplet/gradle/play/PlayPublishApkTask.groovy#L45). The PR adds a new flag onto the command line, allowing to updates lower tracks to the version codes being pushed.

For instance:

> - I toggle `--untrack-old`
>   - Both alpha and beta tracks are at version codes `11` and `12`.  I upload APKs with version `21` and `22` on the **beta** track => alpha gets updated as well.
>   - Alpha and beta tracks are still at `11` and `12`.  I upload APKs with version `21` and `22` on the **alpha** track => beta remains untouched.
>   - Beta tracks is at `11` and `12`, but alpha is at `31` and `32`.  I upload APKs with version `21` and `22` on the **beta** track => alpha isn't changed.
> - If `--untrack-old` isn't provided, the behavior remains the same as it used to be without the PR.

In order to preemptively know the version codes , apks are open and `AndroidManifest.xml` is read. Because this xml file is actually a binary one, I imported a [3rd party lib](https://pypi.python.org/pypi/AxmlParserPY). The project doesn't seem to too active and I'm not sure what the binary parsing exactly does. @arroway , would you think this library is okay to with?  
Also, we'll need to use a newer version of that lib (so it'll be compatible with python 3). I opened a PR for that.

Code-wise, I refactor `upload_apks()` to be smaller. I think we can even make this file purely functional, but I didn't want to make too many changes.
I also added a `--no-commit` option in order to make development simpler (no need to comment out the commit lines, anymore)

One last thing: Python 2 compatibility is currently broken. I can bring back the compatibility, but I'm not sure of the advantages.

r? @sylvestre

### DIFF
--- a/mozapkpublisher/apk_metadata.py
+++ b/mozapkpublisher/apk_metadata.py
@@ -1,0 +1,20 @@
+from xml.dom import minidom
+from zipfile import ZipFile
+
+from axmlparserpy.axmlprinter import AXMLPrinter
+
+
+# Even though a class called APK exists in axmlparserpy, it's not compatible with Python 3 (unlike the rest of the
+# library). That's why a simpler version is defined here
+def extract_metadata_from_apk(file_name):
+    with ZipFile(file_name) as apk_file:
+        with apk_file.open('AndroidManifest.xml') as manifest_file:
+            manifest_binary_content = manifest_file.read()
+
+    manifest_string_content = AXMLPrinter(manifest_binary_content).getBuff()
+    xml_manifest = minidom.parseString(manifest_string_content)
+
+    return {
+        # Google Play uses integers
+        'version_code': int(xml_manifest.documentElement.getAttribute('android:versionCode')),
+    }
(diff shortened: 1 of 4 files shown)


JohanLorenzo:
Unlike I thought last week, a better solution would just to make the staging instance not commit any APK. [Bug 1314063](https://bugzilla.mozilla.org/show_bug.cgi?id=1314063) showed us how staging instance data can break production tracks.

As a consequence, I'd like to revisit the patch, by just exposing the ability to not commit on anything. Clearing review for now.

 ~~r? @sylvestre~~

JohanLorenzo:
After giving it another round of thoughts, I don't think we need `--untrack-old` for now. `--no-commit` is a way better safeguard. The latter option is done in #6. Closing the current PR for now. Feel free to reopen anytime.

PR #3, opened 2016-10-27
Add tests and make them run on CI

JohanLorenzo:


JohanLorenzo:
@sylvestre noticed the division by 100 doesn't work on Python 3. This PR is the perfect place to add a test for it.

JohanLorenzo:
Follow up of #6 => We now have unit tests across the whole project.

The few bits missing are related to `get_apk.py` and `googleplay.py`. In the first case, things related to IO (as in "downloads and cleanups") are not tested yet. I don't believe this is a high priority as `pushapkworker` relies on `scriptworker` to perform downloads (including chain of trust). 
In the second case, I don't think unit tests are a smart way to proceed. The functions not unit tested in `googleplay.py` are one-liners (if we don't take logs into accounts) that directly call the googleplay client. Hence, I think it'd be better to have an integration test that performs a big dry-run. I'd like to handle that in a follow up. The current PR is already big enough (and I apologize for that)

Finally, I decided to go with python 2 for the coveralls reports. We sometimes have try/except imports, to be python 2.7 compatible. As a consequence, the python 3 report shows some missing lines, here and there.

r? @lundjordan

mozapkpublisher/push_apk.py
```diff
@@ -54,111 +43,77 @@ def _init_parser(cls):
...
-                            (apk_response['versionCode'], apk_file_name, edit_id))
-
-                versions.append(apk_response['versionCode'])
-
-                if 'aurora' in self.config.package_name:
-                    logger.warning('Aurora is not supported by the L10n Store (see \
-https://github.com/mozilla-l10n/stores_l10n/issues/71). Skipping what\'s new.')
-                else:
-                    self._push_whats_new(package_code, service, edit_id, apk_response)
+            apk_response = edit_service.upload_apk(apk_file_name)
```
lundjordan: what happens if we fail to upload apk? does .execute() fatal out for us or should we catch key errors like 'versionCode' missing?

mozapkpublisher/push_apk.py
```diff
@@ -54,111 +43,77 @@ def _init_parser(cls):
...
-
-    def _commit_if_needed(self, service, edit_id):
-        if self.config.dry_run:
-            logger.warn('Dry run option was given, transaction not committed.')
-        else:
-            service.edits().commit(
-                editId=edit_id, packageName=self.config.package_name
-            ).execute()
-            logger.info('Changes committed')
+            upload_body[u'userFraction'] = self.config.rollout_percentage / 100.0
```
lundjordan: hehe, I remember this!

mozapkpublisher/push_apk.py
JohanLorenzo: Based on [the documentation](https://developers.google.com/resources/api-libraries/documentation/androidpublisher/v2/python/latest/androidpublisher_v2.edits.apks.html#upload), it's unclear. I tried locally to upload an APK that was already uploaded. `.execute()` failed this way:
```
2016-11-22 10:29:16,844 - http.py - WARNING - Encountered 403 Forbidden with reason "apkUpgradeVersionConflict"
Traceback (most recent call last):
  File "../mozapkpublisher/push_apk.py", line 119, in <module>
    main(__name__)
  File "../mozapkpublisher/push_apk.py", line 112, in main
    PushAPK().run()
  File "../mozapkpublisher/push_apk.py", line 84, in run
    self.upload_apks(apks)
  File "../mozapkpublisher/push_apk.py", line 65, in upload_apks
    apk_response = edit_service.upload_apk(apk_file_name)
  File "/home/jlorenzo/git/mozilla-releng/mozapkpublisher/mozapkpublisher/googleplay.py", line 68, in _transaction_required
    return method(*args, **kwargs)
  File "/home/jlorenzo/git/mozilla-releng/mozapkpublisher/mozapkpublisher/googleplay.py", line 88, in upload_apk
    media_body=apk_path
  File "/home/jlorenzo/.virtualenvs/mozapkpublisher/lib/python3.5/site-packages/oauth2client-3.0.0-py3.5.egg/oauth2client/util.py", line 137, in positional_wrapper
  File "/home/jlorenzo/.virtualenvs/mozapkpublisher/lib/python3.5/site-packages/google_api_python_client-1.5.3-py3.5.egg/googleapiclient/http.py", line 838, in execute
googleapiclient.errors.HttpError: <HttpError 403 when requesting https://www.googleapis.com/upload/androidpublisher/v2/applications/org.mozilla.fennec_aurora/edits/16704049670753158544/apks?alt=json&uploadType=media returned "APK specifies a version code that has already been used.">
```

I believe it's safe enough, so we won't get non obvious errors like `KeyError: 'versionCode'`

JohanLorenzo:
In addition to unit tests, I tested the refactored scripts locally. I believe 1 reviewer is okay, based on the amount of code changed in the non-test code.

PR #4, opened 2016-11-01
Mute 3rd-party logs

JohanLorenzo:


JohanLorenzo:
Travis failed because it shouldn't have run in the first place. Tests are not added yet. They will in #3

PR #5, opened 2016-11-02
Add check to prevent single locales from being pushed

JohanLorenzo:


mozapkpublisher/apk.py
```diff
@@ -0,0 +1,43 @@
...
+
+logger = logging.getLogger(__name__)
+
+LOCALE_LINE_PATTERN = re.compile(r'^locale \S+ (\S+) .+')
+OMNI_JA_LOCATION = 'assets/omni.ja'
+CHROME_MANIFEST_LOCATION = 'chrome/chrome.manifest'
+
+
+def check_if_apk_is_multilocale(apk_path):
+    with ZipFile(apk_path) as apk:
```
flodolo: You can just use apk_zip here, instead of assigning it later

JohanLorenzo: Good point. That was a leftover of a previous approach. Fixed.

sylvestre:
Let's do it

PR #6, opened 2016-11-03
Add --dry-run option

JohanLorenzo:
r? @sylvestre

### DIFF
--- a/mozapkpublisher/base.py
+++ b/mozapkpublisher/base.py
@@ -17,7 +17,20 @@ class Base(object):
 
     @staticmethod
     def _convert_dict_into_args(dict_):
-        dash_dash_dict = {'--{}'.format(key.replace('_', '-')): value for key, value in dict_.items()}
-        flatten_args = [item for tuples in dash_dash_dict.items() for item in tuples]
+        # For instance "dry_run" being True means the argument should be added to the command line.
+        dict_without_deactivated_unary_arguments = {key: value for key, value in dict_.items() if value is not False}
+
+        dash_dash_dict = {
+            '--{}'.format(key.replace('_', '-')): value
+            for key, value in dict_without_deactivated_unary_arguments.items()
+        }
+
+        args_with_unary_arguments_alone = [
+            (key, value) if not isinstance(value, bool) else (key,)
+            for key, value in dash_dash_dict.items()
+        ]
+
+        flatten_args = [item for tuples in args_with_unary_arguments_alone for item in tuples]
+
         logger.debug('dict_ converveted into these args: %s', flatten_args)
         return flatten_args
--- a/mozapkpublisher/push_apk.py
+++ b/mozapkpublisher/push_apk.py
@@ -43,8 +43,7 @@ class PushAPK(Base):
 
         googleplay.add_general_google_play_arguments(cls.parser)
 
-        cls.parser.add_argument('--track', choices=googleplay.TRACK_VALUES,
-                                default='alpha',    # We are not using alpha but we default to it to avoid mistake
+        cls.parser.add_argument('--track', choices=googleplay.TRACK_VALUES, default='alpha',
                                 help='Track on which to upload')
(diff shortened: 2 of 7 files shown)


JohanLorenzo:
Talked to Sylvestre, he's a bit busy these days. ~~r? @sylvestre~~ 

Would you mind taking a look @lundjordan ? This PR basically adds a new option to push_apk.py (the one that [was present in mozharness](https://bugzilla.mozilla.org/show_bug.cgi?id=1305390)). Having it would prevent bad data (coming from staging/dev) to contaminate the production (and only) instance of Google Play Store. A real life example of such a scenario is: [Bug 1314063](https://bugzilla.mozilla.org/show_bug.cgi?id=1314063)

mozapkpublisher/push_apk.py
```diff
@@ -42,8 +42,11 @@ def _init_parser(cls):
 
         googleplay.add_general_google_play_arguments(cls.parser)
 
-        cls.parser.add_argument('--track', choices=googleplay.TRACK_VALUES,
-                                default='alpha',    # We are not using alpha but we default to it to avoid mistake
+        cls.parser.add_argument('--no-commit', dest='should_commit', action='store_false', default=True,
+                                help='''Perform every operation of the transation, except committing. No data will be
+stored on Google Play. Use this option if you want to test the script with the same data more than once.''')
```
lundjordan: I'm don't know what other options are part of the script transaction but I wonder if we should call this option --dry-run given your description: "Use this option if you want to test the script with the same data more than once"

I suppose it depends on if you forsee other 'production' affecting logic being added to this script and if we want to have a blanket 'dry_run' option to ignore all those parts.

Also depends on how explicit you want to be about what we are not running :)

mozapkpublisher/base.py
```diff
@@ -17,7 +17,20 @@ def _parse_config(cls, config=None):
 
     @staticmethod
     def _convert_dict_into_args(dict_):
```
lundjordan: this seems pretty wild. OOC, is this some sort of backwards compatibility tactic to make a config become arguments only to become a config again?

mozapkpublisher/base.py
```diff
@@ -17,7 +17,20 @@ def _parse_config(cls, config=None):
...
+            '--{}'.format(key.replace('_', '-')): value
+            for key, value in dict_without_deactivated_unary_arguments.items()
+        }
+
+        args_with_unary_arguments_alone = [
+            (key, value) if not isinstance(value, bool) else (key,)
+            for key, value in dash_dash_dict.items()
+        ]
+
+        flatten_args = [item for tuples in args_with_unary_arguments_alone for item in tuples]
```
lundjordan: wowza, that's a lot of work to support unary args! :) When is a config (dict) actually passed? Could we just change these configs to hold items that would be the result of `parse_args(args)`? That way, we either parse args or we use a dict, but both yield the same config

mozapkpublisher/base.py
lundjordan: I suppose I sort of ask this question again in the comment below. feel free to ignore this comment :)

mozapkpublisher/base.py
JohanLorenzo: Dict are passed [in pushapkworker](https://github.com/mozilla-releng/pushapkscript/blob/4794e9703e1f534f0afe55300639a1c376215242/pushapkscript/googleplay.py#L11). I didn't want the worker to shell out.

Storing the result of `parse_args()` would make things clearer, indeed. We would loose the input verification, though. At the moment, if the dict contains an extra parameter, or doesn't have an expected value (for instance a string, instead of an int within the range [0-100]), the exception `WrongArgumentGiven` is raised.

I'm not sure what's the best scenario. I initially tended to choose input verification, but I agree, `_convert_dict_into_args()` is becoming a bit of a headache.

JohanLorenzo:
I renamed the flag and rebased in top of master.

PR #7, opened 2016-11-07
Fix a bug in the logger call

sylvestre:


JohanLorenzo:
Like we say, in French : Bien vu !

Thanks

PR #8, opened 2016-11-22
Add troubleshooting tips for Mac OSX

gerrycfchang:
As titled.


JohanLorenzo:
Wow, I didn't know the python libs were relying that much on openssl. Openssl is definitely a pain to handle in Mac OS.

Thank you for documenting that out!

PR #10, opened 2017-01-03
Update API links to stores_l10n

flodolo:


flodolo:
Changes:
* API is now versioned (currently at v1). If version is not provided, it falls back to v1.
* google is ok for query related to stores (e.g. mapping), fx_android should be used when the query is about Firefox for Android (in order to support more products in the future).

flodolo:
Clearly I need to look into tests…

JohanLorenzo:
The changes make sense to me. I'll tell you once it's deployed in production.

PR #11, opened 2017-01-05
Fix what's new message not appearing in the logs

JohanLorenzo:
Thanks for spotting it @sylvestre ! r? 

I tested the patch manually with both python 2 and 3. I got the warning message in one case but not the other. Like said in that message, I also got some characters stripped

### DIFF
--- a/mozapkpublisher/base.py
+++ b/mozapkpublisher/base.py
@@ -35,7 +35,7 @@ class Base(object):
 
         flatten_args = [str(item) for tuples in args_with_unary_arguments_alone for item in tuples]
 
-        logger.debug('dict_ converveted into these args: %s', flatten_args)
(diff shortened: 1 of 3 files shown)


mozapkpublisher/base.py
```diff
@@ -35,7 +35,7 @@ def _convert_dict_into_args(dict_):
 
         flatten_args = [str(item) for tuples in args_with_unary_arguments_alone for item in tuples]
 
-        logger.debug('dict_ converveted into these args: %s', flatten_args)
+        logger.debug('dict_ converveted into these args: {}'.format(flatten_args))
```
flodolo: Typo: converted

JohanLorenzo:
Thanks for making the string conversion simple @flodolo ! r? 

~~r? @sylvestre~~

PR #19, opened 2017-03-10
get_apk: download() doesn't rename files anymore

JohanLorenzo:
~~Needs some integration tests~~

0.2.0 incoming!

### DIFF
--- a/mozapkpublisher/get_apk.py
+++ b/mozapkpublisher/get_apk.py
@@ -33,8 +33,6 @@ class GetAPK(Base):
         )
 
         exclusive_group = cls.parser.add_mutually_exclusive_group(required=True)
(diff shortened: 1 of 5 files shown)


mozapkpublisher/get_apk.py
```diff
@@ -61,88 +58,41 @@ def signal_handler(self, signal, frame):
 
     def cleanup(self):
```
escapewindow: It may be useful to run this before we download, but the current behavior (not dying if it exists and overwriting) is probably acceptable.

JohanLorenzo: Good thinking. I created [this good first bug](https://github.com/mozilla-releng/mozapkpublisher/issues/20).

PR #22, opened 2017-03-24
Bug 1349039 - Add listings of new locales during push_apk

JohanLorenzo:
The patch could have been smaller. We technically just needed to call `edit_service.update_listings()` in `_push_whats_new`. However, I wanted to be sure the data structures passed to googleplay.py were the same as in `update_apk_description.py` (that's the script I used yesterday to unlock the locales).

Hence, I first added more tests to googleplay.py to freeze the calls sent to google-api-python-client. Then, I changed the method parameters in EditService. Finally, the calls are made from push_apk.

I've been cautious because we don't have a way to manually test it, other than when it's in production.

cc @flodolo

### DIFF
--- a/mozapkpublisher/googleplay.py
+++ b/mozapkpublisher/googleplay.py
@@ -108,16 +108,14 @@ class EditService(object):
         logger.debug('Listing body: {}'.format(body))
 
     @transaction_required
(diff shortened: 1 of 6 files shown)


JohanLorenzo:
Definitely :)

PR #24, opened 2017-04-19
Error out if using --rollout-percentage on a non-rollout track

jcristau:
Might avoid accidentally pushing an update to 100% instead of 10%.


JohanLorenzo:
Thank you very much!

PR #29, opened 2017-05-18
get_apk/checksum: Allow files in subdirectories

JohanLorenzo:
While working on https://github.com/escapewindow/staging-release-upload/pull/1, I realized `_fetch_checksum_from_file` doesn't support paths like `update/target.mar` in the checksum file.

This fixes it.

### DIFF
--- a/mozapkpublisher/get_apk.py
+++ b/mozapkpublisher/get_apk.py
@@ -159,10 +159,10 @@ def check_apk_against_checksum_file(apk_file, checksum_file):
 
 
 def _fetch_checksum_from_file(checksum_file, apk_file):
-    base_apk_filename = os.path.basename(apk_file)
+    base_apk_filepath = _take_out_common_path(checksum_file, apk_file)
     with open(checksum_file, 'r') as fh:
         for line in fh:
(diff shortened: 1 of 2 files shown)


mozapkpublisher/get_apk.py
```diff
@@ -175,6 +175,18 @@ def _fetch_checksum_from_file(checksum_file, apk_file):
...
+def _take_out_common_path(checksum_file, apk_file):
+    try:
+        common_prefix = os.path.commonpath((checksum_file, apk_file))  # Only exists in Python 3.5+
+    except AttributeError:
+        common_prefix = os.path.dirname(checksum_file)
+
+    if common_prefix[-1] != '/':
+        common_prefix = '{}/'.format(common_prefix)
+
+    return apk_file.replace(common_prefix, '')
```
escapewindow: This could be `os.path.relpath`, which wouldn't require adding the trailing slash.

```
>>> os.path.relpath("/foo/bar/baz/blah", "/foo/bar")
'baz/blah'
```

This function could therefore be replaced with

`os.path.relpath(apk_file, os.path.dirname(checksum_file))`

but I'm ok with the helper function if wanted.

JohanLorenzo: Much better! I used your solution and kept the helper function, which helps in testing and give a bit more of context.

PR #31, opened 2017-05-29
update_apk_description: Support Firefox Focus

JohanLorenzo:
I couldn't not fully test against `org.mozilla.focus` because:

a. I don't have the rights to modify this app
b. No APK has been uploaded yet. Hence Google Play returns: `No application was found for the given package name.`

The package name was taken from https://github.com/mozilla-mobile/focus-android/blob/09784e886254ae6a398b7156930ebd7efeb2dd91/app/src/main/java/org/mozilla/focus/FocusApplication.java#L6


sylvestre:
Sebastian told me that, we will need to create a separate product for Germany called Klar
we can probably address that into a follow up PR

JohanLorenzo:
A follow up sounds good to me!

PR #33, opened 2017-06-08
Support of klar & Update README.md (minor changes)

sylvestre:


README.md
```diff
@@ -5,6 +5,8 @@ Scripts to publish Firefox for Android on Google Play Store.
 ## Setup and run
 
 1. Create a virtualenv and source it
+   `virtualenv venv`
```
JohanLorenzo: Nit: Use ```, otherwise both commands will be on the same line

PR #40, opened 2017-10-17
push_apk.py: Add option to not update listings and "what's new"

### LINKED ISSUE #36: push_apk.py: Add option to not update listings and "what's new"
[Bug 1383642](https://bugzilla.mozilla.org/show_bug.cgi?id=1383642) caused 55.0b12 to *not* be uploaded. This was a little bit silly because almost no strings had to be updated on Google Play.

To workaound this, folks had to [comment out these lines](https://github.com/mozilla-releng/mozapkpublisher/blob/aa1ea09b88b0f505e28149e1e48811840dd66930/mozapkpublisher/push_apk.py#L59-L64). We can just guard these lines with a condition which depends on a command line argument.

cc @MihaiTabara @jcristau

JohanLorenzo:
Fixes #36 

This came up as part of releaseduty work. I'll handle the changes in the process. Would you be okay to review @MihaiTabara ? If you don't think you have the bandwidth, please let me know. I'll ask someone in releaseduty.

### DIFF
--- a/mozapkpublisher/push_apk.py
+++ b/mozapkpublisher/push_apk.py
@@ -47,6 +47,9 @@ class PushAPK(Base):
         cls.parser.add_argument('--apk-armv7-v15', dest='apk_file_armv7_v15', type=argparse.FileType(),
                                 help='The path to the ARM v7 API v15 APK file', required=True)
 
+        cls.parser.add_argument('--no-gp-string-update', dest='update_google_play_strings', action='store_false',
+                                help="Don't update listings and what's new sections on Google Play")
+
     def upload_apks(self, apks):
         for architecture, apk in apks.items():
             apk_helper.check_if_apk_has_claimed_architecture(apk['file'], architecture)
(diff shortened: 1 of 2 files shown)


JohanLorenzo:
Test failure will be fixed in #42

mozapkpublisher/push_apk.py
```diff
@@ -47,6 +47,9 @@ def _init_parser(cls):
         cls.parser.add_argument('--apk-armv7-v15', dest='apk_file_armv7_v15', type=argparse.FileType(),
                                 help='The path to the ARM v7 API v15 APK file', required=True)
 
+        cls.parser.add_argument('--no-gp-string-update', dest='update_google_play_strings', action='store_false',
```
MihaiTabara: I'm a bit confused. I don't see the correlation between the "negative" argument (no-something) and the destination variable (update-something). All the code looks good but it sounds like the cmd line argument should rather be "gp-string-update", no? 

As in, if I pass the "gp-string-update" argument to the script, it will perform this step as well. Otherwise, it will default to `False`.

JohanLorenzo: Clarified over IRC. The confusion was at `action='store_false'`.

If `--no-gp-string-update` is set, then `update_google_play_strings` is `False`. If unset, then `update_google_play_strings` is `True`. That's because of `"store_false"` https://docs.python.org/2/library/argparse.html#action

@MihaiTabara was okay to leave this variable as is.

PR #41, opened 2017-10-17
 push_apk.py: Pass listings/whatsnew via on-disk files 

### LINKED ISSUE #37: push_apk.py: Pass listings/whatsnew via on-disk files
If https://github.com/mozilla-releng/pushapkscript/issues/25 is possible, then mozapkpublisher has to get strings from json files which are present on the file system.

cc @MihaiTabara

JohanLorenzo:
Fixes #37. Depends on #40. 

The content of the file isn't checked yet. I plan to do so in a follow up patch.

### DIFF
--- a/mozapkpublisher/push_apk.py
+++ b/mozapkpublisher/push_apk.py
@@ -1,8 +1,9 @@
 #!/usr/bin/env python
 
-import sys
 import argparse
+import json
 import logging
+import sys
 
 from mozapkpublisher import googleplay, store_l10n
 from mozapkpublisher import apk as apk_helper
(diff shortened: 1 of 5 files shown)


mozapkpublisher/push_apk.py
```diff
@@ -47,10 +48,18 @@ def _init_parser(cls):
...
-
-    def upload_apks(self, apks):
+        google_play_strings_group = cls.parser.add_mutually_exclusive_group(required=True)
+        google_play_strings_group.add_argument('--no-gp-string-update', dest='update_google_play_strings', action='store_false',
+                                               help="Don't update listings and what's new sections on Google Play")
+        google_play_strings_group.add_argument('--update-gp-strings-from-l10n-store', dest='update_google_play_strings_from_store',
+                                               action='store_true',
+                                               help="Download listings and what's new sections from the l10n store and use them \
+                                               to update Google Play")
+        google_play_strings_group.add_argument('--update-gp-strings-from-file', dest='google_play_strings_file', type=argparse.FileType(),
```
MihaiTabara: Nit: for consistency I'd rename `google_play_strings_file` with `update_google_play_strings_from_file` or alike.

JohanLorenzo: Thanks for the review. I'm not a huge fan of `update_google_play_strings_from_file` because thanks to `type=argparse.FileType()`, we're sure the parameter will be a file. That's why I named it simply `google_play_strings_file`.

Would you have another suggested name?

MihaiTabara: That makes sense; I was more concerned of the prepending `update` to be consistent with the two other mutually exclusive options: `update_google_play_strings` and `update_google_play_strings_from_store`. 
`

JohanLorenzo: I understand. The 2 others are boolean values. Maybe I should name them `must_not_update_google_play_strings` and `must_update_google_play_strings_from_store`, instead?

MihaiTabara: Neah, I think it's all good. I didn't get it in the first run, but they make sense now 👍

JohanLorenzo: Alright :)

PR #44, opened 2017-10-19
Create get_l10n_strings.py script

JohanLorenzo:
Depends on #43

PR #45, opened 2017-10-19
Drop Python 2 support 

JohanLorenzo:
Depends on #44

Keeping the Python 2 support is risky because of UTF-8 strings provided by https://l10n.mozilla-community.org/stores_l10n/.

This hasn't been a big issue until #41. Before then, @jcristau noted that non-ASCII languages didn't display properly, but they've been uploaded correctly.

#41 now reads UTF-8 strings from a JSON file. This requires some decoding, just like in  https://github.com/mozilla-releng/mozapkpublisher/pull/44/commits/f430c6d056e5323fd1fba75dbc53c09c6f77a63a#diff-7c3a9f88ff9e7a9a229e7fdda8bfe7a6R24. Sadly, decoding UTF-8 value back to ASCII doesn't preserve the UTF-8 characters.

Then, we should stop supporting Python 2, and make explicit that people need Python 3. Otherwise, somebody may end up uploading broken strings to Google Play.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,7 @@ Scripts to publish Firefox for Android on Google Play Store.
 
 ## Setup and run
 
+1. :warning: You need Python >= 3.5 to run these set of scripts. Python 2 isn't supported starting version 0.5.0
 1. Create a virtualenv and source it
 ```sh
 virtualenv venv
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ Scripts to publish Firefox for Android on Google Play Store.
 
 ## Setup and run
 
-1. :warning: You need Python >= 3.5 to run these set of scripts. Python 2 isn't supported starting version 0.5.0
+1. :warning: You need Python >= 3.5 to run this set of scripts. Python 2 isn't supported starting version 0.5.0
 1. Create a virtualenv and source it
 ```sh
 virtualenv venv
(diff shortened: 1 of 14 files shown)


JohanLorenzo:
Review suggestion: Just look at the last commit. The other commits are from other PRs.

JohanLorenzo:
Previous PRs have made master. Diff is now nicer to review.

README.md
```diff
@@ -4,6 +4,7 @@ Scripts to publish Firefox for Android on Google Play Store.
 
 ## Setup and run
 
+1. :warning: You need Python >= 3.5 to run these set of scripts. Python 2 isn't supported starting version 0.5.0
```
jcristau: s/these/this/ or s/set of//

mozapkpublisher/test/integration/test_get_l10n_strings.py
```diff
@@ -12,14 +12,13 @@
 @pytest.mark.skipif(strtobool(os.environ.get('SKIP_NETWORK_TESTS', 'true')), reason='Tests requiring network are skipped')
 @pytest.mark.parametrize('package_name', STORE_PRODUCT_DETAILS_PER_PACKAGE_NAME.keys())
 def test_download_files(package_name):
-    with tempfile.NamedTemporaryFile() as f:
+    with tempfile.NamedTemporaryFile('w+t') as f:
```
jcristau: encoding='utf-8'?

JohanLorenzo: Good catch. Done!

README.md
JohanLorenzo: Done

PR #38, opened 2017-08-18
Bug 1391622 - Add more checks to APKs by splitting data extraction from analysis

JohanLorenzo:
~~**WIP**~~

~~Tox is broken because of the latest version of androguard not being on Pypi. That's tracked at https://github.com/androguard/androguard/issues/236~~

~~Thus, tests haven't been added yet. This will be done before PR is merged.~~

~~This also requires some manual testing on Python 2 (was developed under Python 3).~~

PR #49, opened 2018-04-17
 We moved to api 16 (from api 15), updating the download script

sylvestre:
Still fails on checksums


JohanLorenzo:
One integration test will fail, because of this change. We may want to define the API level based on the major version, just like it's being implemented at https://github.com/mozilla-releng/mozapkpublisher/pull/38/files#diff-28879601bcb073e9d7e0356680d74ea9R18. @sylvestre, when do you want the fix landed? I can implement something more robust by the end of next week.

sylvestre:
When you have time, I downloaded the apk by hands!
so, no rush at all

JohanLorenzo:
I have an more long term solution in #58. Let's fix this for real.

PR #54, opened 2018-04-30
Closes #53 - Remove unneeded check for distinct package names.

### LINKED ISSUE #53: Can't upload Klar with geckoview: "APKs share Package name"
Follow up #52. It turns out one of the checks[1] written is not valid anymore[2] if we ship Klar geckoview alongside Klar webview. Would you mind taking a look @liuche ? 

[1] https://github.com/mozilla-releng/mozapkpublisher/blob/8d4b63cfe48a4b0088a9cfce3a97e4b535ae5bf6/mozapkpublisher/common/apk/checker.py#L78
[2] https://tools.taskcluster.net/groups/Eu2y9DeETMK75GkFDBKJ6w/tasks/PcggY5apRSWwcokoFguxfw/runs/1/logs/public%2Flogs%2Flive_backing.log#L21

liuche:
Removing the check because it's overly specific now, and redundant because we check for max number of package names.

### DIFF
--- a/mozapkpublisher/common/apk/checker.py
+++ b/mozapkpublisher/common/apk/checker.py
@@ -75,7 +75,6 @@ def cross_check_fennec_apks(apks_metadata_per_paths):
 
 
 def cross_check_focus_apks(apks_metadata_per_paths):
-    _check_piece_of_metadata_is_distinct('package_name', 'Package name', apks_metadata_per_paths)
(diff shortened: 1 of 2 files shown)


JohanLorenzo:
That's now deployed. We'll be able to see it in next nightly at noon UTC.

Issue #67, opened 2018-06-25
port to v3 of the google play developer API


jcristau:
We're using v2 at the moment, should probably look at porting to v3.
Doc for v3 is at https://developers.google.com/android-publisher/#publishing, v2 at https://developers.google.com/android-publisher/archive/v2/

lundjordan:
@JohanLorenzo - is this something we should prioritize?

@jcristau - what do we gain? What happens if we don't do it?

JohanLorenzo:
I'm not aware of a planned deprecation of the v2 API. I'd say: let move to v3 (or above) whenever v2 is deprecated or whenever we need some bits of data that is not exposed by the v2 API. How does that sound?

lundjordan:
makes sense. wfm

JohanLorenzo:
> I'm not aware of a planned deprecation of the v2 API.

It turns out there was this blog post[1], which was released 3 days before my comment 😅 

API v2 will be decommissioned on December 1st, 2019.

[1] https://android-developers.googleblog.com/2019/03/changes-to-google-play-developer-api.html

PR #68, opened 2018-06-25
Add a script to check whether a release stays on staged rollout for t…

jcristau:
…oo long

We've had a few cases of a release being staged at 99% and forgetting to
finish the rollout by bumping it to 100%, then having that release
overwritten by the next one at 5 or 10%.  Let's try to avoid that by
checking for a 7-day-old release on staged rollout and nagging.

Ideally this would use mozapkpublisher.common.googleplay but that would require use of v3 API (#67) so this is standalone for now.

### DIFF
--- a/mozapkpublisher/check_rollout.py
+++ b/mozapkpublisher/check_rollout.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python3
+
+import argparse
+import calendar
+import email.utils as eu
+import logging
+import time
+
+import requests
+
+from mozapkpublisher.common import googleplay
+
+DAY = 24 * 60 * 60
+
+logger = logging.getLogger(__name__)
+
+
+def check_rollout(edits_service, package_name, days):
+    """Check if package_name has a release on staged rollout for too long"""
+    edit = edits_service.insert(body={}, packageName=package_name).execute()
+    response = edits_service.tracks().get(editId=edit['id'], track='production', packageName=package_name).execute()
+    releases = response['releases']
+    for release in releases:
+        if release['status'] == 'inProgress':
+            url = 'https://archive.mozilla.org/pub/mobile/releases/{}/SHA512SUMS'.format(release['name'])
+            resp = requests.head(url)
+            if resp.status_code != 200:
+                if resp.status_code != 404:  # 404 is expected for release candidates
+                    logger.warning("Could not check %s: %s", url, resp.status_code)
+                continue
+            age = time.time() - calendar.timegm(eu.parsedate(resp.headers['Last-Modified']))
+            if age >= days * DAY:
+                yield release, age
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Check for in-progress Firefox for Android staged rollout')
+    parser.add_argument('service_account', help='The service account email')
+    parser.add_argument('credentials', help='The p12 authentication file', type=argparse.FileType(mode='rb'))
+    parser.add_argument('--days', help='The time before we warn about incomplete staged rollout of a release (default: 7)',
+                        type=int, default=7)
+    config = parser.parse_args()
+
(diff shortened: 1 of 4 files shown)


mozapkpublisher/check-rollout.py
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/python3
```
JohanLorenzo: Nit: `#!/usr/bin/env python3`
Nit: `check_rollout.py` for consistency across files

mozapkpublisher/check-rollout.py
```diff
@@ -0,0 +1,57 @@
...
+
+import requests
+
+from oauth2client.service_account import ServiceAccountCredentials
+from apiclient.discovery import build
+
+DAY = 24 * 60 * 60
+
+
+def _connect(service_account, credentials_path):
```
JohanLorenzo: This is almost a copy of `googleplay._connect()`. Could we expose that function (in the `googleplay` module) and make the version number a function parameter?

mozapkpublisher/check-rollout.py
```diff
@@ -0,0 +1,57 @@
...
+    """Connect to v3 of the google play store publisher API"""
+    scope = 'https://www.googleapis.com/auth/androidpublisher'
+    credentials = ServiceAccountCredentials.from_p12_keyfile(service_account, credentials_path, scopes=scope)
+    http = httplib2.Http()
+    http = credentials.authorize(http)
+    service = build('androidpublisher', 'v3', http=http, cache_discovery=False)
+    return service
+
+
+def check_rollout(edits_service, package_name, days):
```
JohanLorenzo: Nit: Could you add a `TODO` to say this function should be moved to `googleplay.EditService` when the whole class is ported to the v3 API?

mozapkpublisher/check-rollout.py
```diff
@@ -0,0 +1,57 @@
...
+    """Check if package_name has a release on staged rollout for too long"""
+    edit = edits_service.insert(body={}, packageName=package_name).execute()
+    response = edits_service.tracks().get(editId=edit['id'], track='production', packageName=package_name).execute()
+    releases = response['releases']
+    for release in releases:
+        if release['status'] == 'inProgress':
+            url = 'https://archive.mozilla.org/pub/mobile/releases/{}/SHA512SUMS'.format(release['name'])
+            resp = requests.head(url)
+            if resp.status_code != 200:
+                continue
```
JohanLorenzo: Nit: We might want to warn on errors. Could you add a `log.warn()` call?

mozapkpublisher/check-rollout.py
```diff
@@ -0,0 +1,57 @@
...
+            age = time.time() - time.mktime(parsedate(resp.headers['Last-Modified']))
```
JohanLorenzo: :warning: Last-Modified is in GMT (from the CDN I get) but mktime() makes it localtime https://docs.python.org/3/library/time.html#time.mktime. Just to ensure we don't break timezone, could you make both operand in UTC?

mozapkpublisher/check-rollout.py
```diff
@@ -0,0 +1,57 @@
...
+            if age >= 7 * DAY:
```
JohanLorenzo: Nit: this logic is not DST-proof. I'm fine with that. I'd make the help text explicit about this, though

JohanLorenzo:
Oh, I didn't realize there were no tests for this module. Could you add some. For this kind of script, you might want to re-use some of the stuff in https://github.com/mozilla-releng/mozapkpublisher/blob/master/mozapkpublisher/test/test_push_apk.py

jcristau:
@JohanLorenzo I pushed an update hopefully addressing all your comments.  Thanks for the review!

PR #76, opened 2018-07-03
Always use known requirements

JohanLorenzo:
Let's have only one requirement file. This one enables reproducible builds.


escapewindow:
The point of having a `-dev` file or a `.in` file is to list the direct dependencies, unpinned. Then we can expand that into a pinned file with all upstream dependencies. I agree having a file that we point to for the source of truth is good, but what happens when a direct dependency drops an upstream dependency? With the `-dev` example, when we regenerate the built, pinned file, the dropped dependency will no longer be listed. If we have a single, expanded file, then we might keep that upstream dependency forever.

Having said that, we have a `-dev` file here with pins, which isn't desired either.

escapewindow:
```
10:21 <aki> jlorenzo: do you think an unpinned -dev or .in file makes sense? https://github.com/mozilla-releng/mozapkpublisher/pull/76
10:24 <jlorenzo> aki: that's what we had before and it caused issues. I kinda like the idea of being notified by pyup, and have the deps tested in travis
10:24 <aki> issues how?
10:25 <jlorenzo> discrepancies of behavior between dev and prod
10:25 <aki> you don't have to point at the .in file for anything but expanding your prod deps
10:26 <aki> the use case is, when you add or remove a direct dependency (i.e., `import` lines), you modify the .in or -dev file. then use pip-compile or dephash to expand it to the prod file. that tool will add new upstream deps and remove old ones
10:27 <aki> make sense? or still against it
10:30 <jlorenzo> aki: ah, I see. If you remove the dep from the .in file, then it's easier to delete its sub-deps, right? 
10:30 <aki> yes
10:30 <aki> whereas it's harder to tell whether you can in the all-in-one req file
10:32 lisa → lisa|mtg
10:36 <jlorenzo> that makes sense. I'll keep it then 
10:39 <Callek> jlorenzo: the unpinned/pinned stuff and pinned deps tested + pyup is in place on addonscript, and I like how I did it there with tox and stuff
10:40 <Callek> basically tox should run against pinned, pyup should update pinned, unpinned should exist for unpinned => pinned sanity (and unpinned should be direct deps only)
10:43 <aki> i agree with that
```

JohanLorenzo:
Okay, we're almost at par with addonscript. A few things:

* I kept coveralls. I haven't used codecov.
* In `.coveragerc` I didn't  turn `branch = True` because of the crash seen in 12b17c4/
* I named files `.in` instead of `-dev`. I think the former is clearer about the intent of the files.

What do you guys think?

.travis.yml
```diff
@@ -1,17 +1,20 @@
...
-python:
-  - "3.6"
-  - "3.5"
+
+matrix:
+  include:
+    - python: 3.5
+      env: TOXENV=py35
+    - python: 3.6
+      env: TOXENV=py36
```
escapewindow: We probably want to add 3.7! But we need to allow it to fail until everything's green.

https://github.com/mozilla-releng/signingscript/blob/master/.travis.yml#L3-L11 is how we're doing it until https://github.com/travis-ci/travis-ci/issues/9815 is resolved, although there's a [workaround](https://github.com/travis-ci/travis-ci/issues/9815#issuecomment-402045581) to use 3.7 instead of 3.7-dev.

tox.ini
```diff
@@ -12,24 +12,24 @@ passenv =
...
     flake8
 
-[testenv:py36-coveralls]
-deps=
-    python-coveralls
-    coverage>=4.2b1
+[testenv:coveralls]
+deps = -rrequirements-coveralls.txt
 commands=
+    pip check
```
Callek: not a blocker, but I discovered in addonscript that pip check here does *not* cause the travis job to fail, because its run 'after_success' so failure doesn't break travis.

JohanLorenzo: Good point. I moved that check in `script` instead. Thanks for raising it up!

PR #111, opened 2018-08-02
Add integration tests for push_apk

JohanLorenzo:
We're still not pushing anything to Google Play in these tests, however we check the other dependencies (like androguard, voluptuous, etc).

That will give more confidence in the updates pyup suggests.


JohanLorenzo:
Closing this PR in favor of https://bugzilla.mozilla.org/show_bug.cgi?id=1575691

PR #78, opened 2018-07-04
Add bulk upload for replies

fiji-flo:
This was part for the SuMo playstore sprint. We used this functionality to bulk upload replies to playstore reviews.

PR #122, opened 2018-09-13
Run tests on Python 3.7

JohanLorenzo:


tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py35, py36
+envlist = py35, py36, py37
```
MihaiTabara: I think we can drop support for 35, can't we? AFAIK we migrated most of systems to 36 in our infra.

JohanLorenzo: I'd prefer to keep it for now. mozapkpublisher is used by other teams on other systems (see https://github.com/mozilla-releng/mozapkpublisher/pull/78 for instance). I don't want to break them without being explicit about it. I did call people out when Python 2 was dropped (https://github.com/mozilla-releng/mozapkpublisher/pull/45).

PR #139, opened 2018-11-30
Adds reference-browser as another focus-esque publishable browser

mitchhentges:


mozapkpublisher/common/utils.py
```diff
@@ -51,4 +52,8 @@ def contains_value(value):
 
     @staticmethod
     def is_focus_flavor(value):
```
mitchhentges: Perhaps there's a better term to encapsulate what we're expecting when we want a `focus_flavor`-y app. Doesn't upload strings to Google Play automatically? Different apk structure (e.g.: no `lib` folder)?
In the mean-time, I just kept the same method name and added `REFERENCE_BROWSER`, but that's kinda sleazy

mozapkpublisher/common/apk/checker.py
```diff
@@ -46,7 +46,7 @@ def cross_check_fennec_apks(apks_metadata_per_paths):
 
 def cross_check_focus_apks(apks_metadata_per_paths):
     _check_number_of_distinct_packages(apks_metadata_per_paths, 2)
-    _check_correct_apk_product_types(apks_metadata_per_paths, [PRODUCT.FOCUS, PRODUCT.KLAR])
+    _check_correct_apk_product_types(apks_metadata_per_paths, [PRODUCT.FOCUS, PRODUCT.KLAR, PRODUCT.REFERENCE_BROWSER])
```
JohanLorenzo: I'm afraid this change will introduce unwanted behavior. If I remember correctly, `_check_correct_apk_product_types()` was implemented to make sure we're pushing 2 different package names (Firefox Focus and Klar are the same product, but for different different markets, because of branding/legal). Therefore, let's create `cross_check_reference_browser_apks()` without this check. Instead, you may use:

* `_check_all_apks_have_the_same_package_name()`
* `_check_apks_version_codes_are_correctly_ordered()` 
* `_check_all_architectures_and_api_levels_are_present()`

mozapkpublisher/common/utils.py
JohanLorenzo: I see! `is_focus_flavor()` was meant to be really restricted to Firefox Focus and its German equivalent Firefox Klar. To me, let's not change this function.

mozapkpublisher/common/apk/checker.py
```diff
@@ -41,13 +46,16 @@ def cross_check_fennec_apks(apks_metadata_per_paths):
 
     _check_all_architectures_and_api_levels_are_present(apks_metadata_per_paths)
 
-    logger.info('APKs are sane!')
+
+def cross_check_reference_browser_apks(apks_metadata_per_paths):
+    _check_all_apks_have_the_same_package_name(apks_metadata_per_paths)
+    _check_apks_version_codes_are_correctly_ordered(apks_metadata_per_paths)
```
JohanLorenzo: Note for the future: The mobile team may want to support ARM64. If that's the case, we should revisit the implementation `_check_apks_version_codes_are_correctly_ordered()`. At the moment, it breaks on purpose because of (bug 1368484 comment 4)[https://bugzilla.mozilla.org/show_bug.cgi?id=1368484#c4]

PR #142, opened 2018-12-14
Removes firefox_* extraction for reference-browser

mitchhentges:
Perhaps the [`check_all_arch_and_api`](https://github.com/mozilla-releng/mozapkpublisher/blob/master/mozapkpublisher/common/apk/checker.py#L53) check should be removed, too :thinking:

### DIFF
--- a/mozapkpublisher/common/apk/extractor.py
+++ b/mozapkpublisher/common/apk/extractor.py
@@ -41,14 +41,16 @@ def extract_metadata(original_apk_path):
         metadata['api_level'] = int(androguard_apk.get_min_sdk_version())
         metadata['version_code'] = androguard_apk.get_androidversion_code()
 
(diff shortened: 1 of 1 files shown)


mozapkpublisher/common/apk/extractor.py
```diff
@@ -46,10 +46,12 @@ def extract_metadata(original_apk_path):
 
         with ZipFile(apk_copy.name) as apk_zip:
             metadata['architecture'] = _extract_architecture(apk_zip, original_apk_path)
-            metadata['firefox_version'] = _extract_firefox_version(apk_zip)
-            metadata['firefox_build_id'] = _extract_firefox_build_id(apk_zip)
             metadata['locales'] = _extract_locales(apk_zip)
 
+            if not PRODUCT.is_reference_browser(package_name):
```
MihaiTabara: Probably we could enrich this to wider products in the future (I'm surprised `focus` didn't hit this for example) but LGTM for now! 👍

PR #152, opened 2019-01-14
Bug 1368484 - Allow AArch64 on Nightly only

JohanLorenzo:
This needs to be landed today, so the team in charge of ARM64 (aka AArch64, aka arm64-v8a) can have stability results early enough. @MihaiTabara would you mind doing a first pass of review? I'll address @mitchhentges 's comments in a follow up PR.


mozapkpublisher/common/apk/history.py
```diff
@@ -36,8 +39,20 @@ def get_expected_api_levels_for_version(firefox_version):
...
 
 
-def get_expected_architectures_for_version(firefox_version):
-    return _get_expected_things_for_version(firefox_version, _MAJOR_FIREFOX_VERSIONS_PER_ARCHITECTURE, 'architecture')
+def get_expected_architectures_for_version(firefox_version, package_name):
+    expected_architectures = list(_get_expected_things_for_version(
+        firefox_version, _MAJOR_FIREFOX_VERSIONS_PER_ARCHITECTURE, 'architecture'
+    ))
+
+    # XXX arm64-v8a (aka AArch64) is not planned to ride trains regularly. It may need a couple of
```
MihaiTabara: 👍 for the comment.

mozapkpublisher/common/apk/history.py
```diff
@@ -36,8 +39,20 @@ def get_expected_api_levels_for_version(firefox_version):
...
+    # cycles to stabilize. That's why we just expect it on Nightly, for now.
+    major_version = get_firefox_major_version_number(firefox_version)
+    first_aarch64_version = _MAJOR_FIREFOX_VERSIONS_PER_ARCHITECTURE['arm64-v8a']['first_firefox_version']
+
+    if major_version >= first_aarch64_version and package_name != 'org.mozilla.fennec_aurora':
```
MihaiTabara: Question for later, not blocking: I'm out of context a bit on how the package names look like in Google Play store, but why is the channel named `aurora` related instead of `nightly` here? 🤔 Was this a side effect from the Dawn project or did we have this from the very beginning and I completely missed this?

JohanLorenzo: You're right, it's a side effect of Dawn. Actually, your comment made me realize I was hardcoding the package name another time. There's a dedicated function, with a comment explaining why nightly is `org.mozilla.fennec_aurora`. I made the change

MihaiTabara: Ah, found it in your https://github.com/mozilla-releng/mozapkpublisher/pull/152/commits/930487dd43a8fb6a96b5593f3648eb5b05775bec - there's https://github.com/mozilla-releng/mozapkpublisher/blob/master/mozapkpublisher/common/apk/checker.py#L105. TIL!

PR #155, opened 2019-01-18
Runs file-creating test in temporary directory

mitchhentges:
Whenever we run our `get_l10n_strings` test, it creates a file in our `mozapkpublisher/test` directory that isn't `gitignore`'d.
I think that if our tests create temporary files, we should use a temporary directory (e.g.: what if we assert on such files in the future? The state should be clean).

This PR ensures that a temp directory is used for the environment-mutating test


JohanLorenzo:
There's a python3.5-specific error. It seems like an API issue. Could you fix it before landing it?

PR #156, opened 2019-01-22
RFC 6: Remove product specific code

mitchhentges:
[RFC](https://github.com/mozilla-releng/releng-rfcs/blob/master/rfcs/0006-pushapkscript-product-to-config.md)

Notes:
* I'm not a fan of all the Fennec checks are grouped under a `--skip-checks-fennec` flag - however, these are _so_ Firefox-specific (e.g. the mapping of version codes to package name), I had two options:
    1. Leave the checks in `mozapkpublisher` and leave them grouped under a physical product, thereby still leaking product information to `mozapkpublisher`
    2. Move these Firefox-specific checks up a level to `pushapkscript` (or maybe even in-tree :thinking:) However, this would be a significant change (most of `extractor.py` and `history.py` are Firefox-specific). This would increase the diff significantly. This is worth investigating later
* `l10n` still has product-specific information, but it's not just called by `pushapkscript`. I'll leave this for future investigation as well
* Most importantly, this change still accomplishes the goal of "when adding a new product, `mozapkpublisher` should not need to be edited"


mozapkpublisher/common/apk/extractor.py
```diff
@@ -57,8 +54,8 @@ def extract_metadata(original_apk_path):
 
 def _extract_architecture(apk_zip, original_apk_path):
     files_with_architecture_in_path = [
-        file_info.filename for file_info in apk_zip.infolist()
-        if _DIRECTORY_WITH_ARCHITECTURE_METADATA in file_info.filename
+        name for name in apk_zip.namelist()
```
JohanLorenzo: Good call. It's simpler!

mozapkpublisher/test/common/apk/test_checker.py
```diff
@@ -1,17 +1,58 @@
...
+    _check_all_apks_have_the_same_firefox_version, _check_version_matches_package_name, \
     _check_all_apks_have_the_same_build_id, _check_all_apks_have_the_same_locales, \
     _check_piece_of_metadata_is_unique, _check_apks_version_codes_are_correctly_ordered, \
     _check_all_apks_are_multi_locales, _check_all_architectures_and_api_levels_are_present
-from mozapkpublisher.common.utils import PRODUCT
 from mozapkpublisher.common.exceptions import NotMultiLocaleApk, BadApk, BadSetOfApks
 
 
-@pytest.mark.parametrize('apks_metadata_per_paths', ({
+def test_check_correct_apk_package_names():
```
JohanLorenzo: Nit: Let's use `@pytest.mark.parametrize()` to split this test into 3.

mozapkpublisher/common/apk/checker.py
```diff
@@ -16,65 +15,40 @@
...
 
-def cross_check_focus_apks(apks_metadata_per_paths):
-    _check_number_of_distinct_packages(apks_metadata_per_paths, 2)
-    _check_correct_apk_product_types(apks_metadata_per_paths, [PRODUCT.FOCUS, PRODUCT.KLAR])
+    if not skip_check_package_names:
+        _check_apk_package_name(apks_metadata_per_paths, expected_package_names)
 
+    logger.info('APKs are sane!')
 
-def _check_number_of_distinct_packages(apks_metadata_per_paths, max_packages):
```
JohanLorenzo: I'm not too sure, does this check have an equivalent?

CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [upcoming]
+
+### Added
+* Options to skip specific sanity checks
+    * `--skip-check-ordered-version-codes`: Skip check that asserts version codes are different, x86 code > arm code
```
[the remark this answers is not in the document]
mitchhentges: [Focus currently doesn't check that version codes are ordered](https://github.com/mozilla-releng/mozapkpublisher/blob/master/mozapkpublisher/common/apk/checker.py#L62-L64)

mozapkpublisher/common/apk/checker.py
mitchhentges: Yeah, I joined this with `_check_apk_package_name`

JohanLorenzo: Okay, great!

CHANGELOG.md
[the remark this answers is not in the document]
JohanLorenzo: Ah yeah, I forgot about this one. I remember now, Klar x86 and Focus x86 have the same version code which broke the check. I'm fine to keep the flag :)

mitchhentges:
I need to remove some additional product-specific properties.
One of them is handled in [this PR](https://github.com/mozilla-releng/pushapkscript/pull/72)

mitchhentges:
Re-pinging for review since it's been a while, there's been merge conflicts, etc

PR #159, opened 2019-01-25
Refactors push_apk to use a variable Google Play Strings interface

mitchhentges:
Let me know if this isn't pythonic.
The benefit of this change is that `PushApk` doesn't need to juggle three parameters for Google Play and carefully judge how they interact: now, it just takes an implementation that has `get_strings(...)`, and it's content.

Let me know what you think :)


mozapkpublisher/push_apk.py
```diff
@@ -14,41 +13,77 @@
 logger = logging.getLogger(__name__)
 
 
-def push_apk(apks, service_account, google_play_credentials_file, track, update_google_play_strings=True,
-             update_google_play_strings_from_store=False, google_play_strings_file=None, rollout_percentage=None,
-             commit=True, contact_google_play=True):
+class NoGooglePlayStrings:
+    @staticmethod
```
mitchhentges: Let me know if this `@staticmethod` adds anything for you.
My IDE was bugging me, and it makes sense since `self` isn't needed, but I'm not sure how weird it is in Python to have the noise of `@staticmethod`.

mozapkpublisher/push_apk.py
```diff
@@ -161,15 +185,19 @@ def main(name=None):
                                            help="Use file to update listing and what's new section on Google Play.\
                                                            Such file can be obtained by calling fetch_l10n_strings.py")
 
+    config = parser.parse_args()
+    if config.update_google_play_strings_from_store:
+        google_play_strings = StoreGooglePlayStrings()
+    elif config.google_play_strings_file:
+        google_play_strings = FileGooglePlayStrings(config.google_play_strings_file)
+    else:
```
JohanLorenzo: Nit:

```py
elif not update_google_play_strings:
    google_play_strings = NoGooglePlayStrings()
else:
    raise WrongArgumentGiven("Option missing. You must provide what to do in regards to Google Play strings.")
```

This prevents undocumented errors. If we want to make this the default behavior, then let's update the documentation.

mozapkpublisher/push_apk.py
JohanLorenzo: Yeah, having `@staticmethod` makes sense to me, here. As far as I know this decorator isn't considered not pythonic.

mozapkpublisher/push_apk.py
mitchhentges: Since the [`google_play_strings_group` is `required=True`](https://github.com/mozilla-releng/mozapkpublisher/pull/159/files/e3af7b1a1be1b09045acf3b452d5eda58be20c9b#diff-45a657627129bc0d2b7ae44372d28670R174), we can be confident that at least one of the options was provided :)

PR #180, opened 2019-04-04
Bug 1520261 - Let ARM64 Fennec ride the trains to Beta and Release

JohanLorenzo:


JohanLorenzo:
⚠️  Please do not merge. I just heard from @pascalchevrel that we may want to ship ARM64 with Fennec 67 beta (maybe not Fennec 67 release). I'll keep this PR on hold while a decision is made.

PR #178, opened 2019-04-01
Bug 1540152 - Create checks_apks.py script 

JohanLorenzo:

PR #182, opened 2019-04-05
check_apks.py: log results out

JohanLorenzo:
Follows #178 up

I found the output of this task very limited 😅  https://tools.taskcluster.net/groups/PjeTZNp1TMO5l40wtL4j5g/tasks/PjeTZNp1TMO5l40wtL4j5g/runs/0/logs/public%2Flogs%2Flive.log#L190

### DIFF
--- a/mozapkpublisher/check_apks.py
+++ b/mozapkpublisher/check_apks.py
@@ -2,6 +2,7 @@
 
 import argparse
 
+from mozapkpublisher.common import main_logging
 from mozapkpublisher.common.apk import add_apk_checks_arguments, extract_and_check_apks_metadata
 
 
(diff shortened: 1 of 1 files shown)


mozapkpublisher/check_apks.py
```diff
@@ -14,6 +15,8 @@ def main():
 
     config = parser.parse_args()
 
+    main_logging.init()
```
mitchhentges: I wonder if this is context-manager-able, so it's harder to miss?
```
with main_logging as logger:
    # do things with logger variable
```
:thinking:

JohanLorenzo: Good idea. I'm not too sure, how we would add it once and for all. I'll think about it!

PR #196, opened 2019-05-27
Bug 1554501 - Do not raise an error when an APK has already been uplo…

JohanLorenzo:
…aded

### DIFF
--- a/mozapkpublisher/common/googleplay.py
+++ b/mozapkpublisher/common/googleplay.py
@@ -14,10 +14,12 @@
 
 import argparse
 import httplib2
+import json
 import logging
 
 from apiclient.discovery import build
 from oauth2client.service_account import ServiceAccountCredentials
+from googleapiclient.errors import HttpError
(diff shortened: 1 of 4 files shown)


mozapkpublisher/common/googleplay.py
```diff
@@ -80,14 +82,23 @@ def commit_transaction(self):
...
     def upload_apk(self, apk_path):
         logger.info('Uploading "{}"'.format(apk_path))
-        response = self._service.apks().upload(
-            editId=self._edit_id,
-            packageName=self._package_name,
-            media_body=apk_path
-        ).execute()
-        logger.info('"{}" uploaded'.format(apk_path))
-        logger.debug('Upload response: {}'.format(response))
-        return response
```
JohanLorenzo: `response` isn't used anywhere but in tests, anymore. It used to when we didn't decode APKs to look at their version codes.

JohanLorenzo:
Failure may be fixed by #197

PR #197, opened 2019-05-27
Remove Python 3.5 support

JohanLorenzo:
In #196, a python3.5-specific error broke these tests: https://travis-ci.org/mozilla-releng/mozapkpublisher/jobs/537876097#L849.

We don't use python 3.5 anymore in production. Scriptworker itself doesn't support this version anymore https://github.com/mozilla-releng/scriptworker/blob/c3953fccc8860073b1f87cf7cc50c89cdedba92d/setup.py#L93. 

On the other side, mozapkpublisher can be run on a local machine and Python 3.5 is supported for another year and a half. I think it's fine to remove this support, based on the usage of this package. What do you think @mitchhentges ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,6 @@ language: python
 
 matrix:
   include:
-    - python: 3.5
-      env:
-        - TOXENV=py35
-        - SKIP_NETWORK_TESTS=0
     - python: 3.6
       env:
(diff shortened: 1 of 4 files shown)


JohanLorenzo:
Awesome! Thanks for your input 🙂

Issue #204, opened 2019-06-20
Disable automatic update of Play Store content


flodolo:
CC @delphine @lwa-moz

Marketing wants to own updating the content on Play Store. In the future we might plug in an external vendor, and need to figure out the API permissions, but for now we want to disable the automatic update from stores_l10n as part of the apk upload.

JohanLorenzo:
Thanks for filing this issue, @flodolo! Some in-tree changes is also needed. I'm going to track this issue in bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1560876

PR #206, opened 2019-06-24
Bug 1560876 - push_apk.py: Stop uploading Google Play strings

JohanLorenzo:
@flodolo and I discussed during the Whistler All Hands. Google Play strings won't be owned by the L10n team anymore, Marketing will. So far, Marketing has updated them via the Web UI. Maybe they'll want to automate it. In the meantime, let's remove the automatic submission. 

If needed, we can bring it back thanks to the Git history. 

See #204 for more context.

### DIFF
--- a/mozapkpublisher/push_apk.py
+++ b/mozapkpublisher/push_apk.py
@@ -1,39 +1,13 @@
 #!/usr/bin/env python3
 
 import argparse
-import json
 import logging
 
-from mozapkpublisher.common import googleplay, store_l10n, main_logging
+from mozapkpublisher.common import googleplay, main_logging
 from mozapkpublisher.common.apk import add_apk_checks_arguments, extract_and_check_apks_metadata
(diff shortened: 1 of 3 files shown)


JohanLorenzo:
Thanks for the quick turnaround Mihai! If @mitchhentges doesn't mind, I'd like to have his input because he implemented some parts of the logic, here.

mozapkpublisher/push_apk.py
```diff
@@ -1,49 +1,22 @@
 #!/usr/bin/env python3
 
 import argparse
-import json
 import logging
 
-from mozapkpublisher.common import googleplay, store_l10n, main_logging
```
mitchhentges: Sounds like we still want the other `store_l10n` functionality for updating apk descriptions and fetching the l10 strings from the store, yeah?

JohanLorenzo: Yeah, let's keep it around, in case someone needs it before it goes away. I'll delete it later.

PR #215, opened 2019-07-15
Top level option abstractions

### LINKED ISSUE #195: Merge "track" and "rollout_percentage"
A track implementation should be provided to the `track` parameter.
This way, when `track` is `rollout`, we can bundle `rollout_percentage` into the configuration object provided. Meanwhile, when track is something else, like `beta`, then `rollout_percentage` isn't needed, and doesn't add noise

mitchhentges:
Fixes #195

This also contains #213, so I wouldn't recommend reviewing until that change lands.

**Improvements:**
* `--service-account` and `--credentials` are now only required if `--do-not-contact-google-play` isn't provided
* `push_apk` and `update_apk_description` now accept types for `track` and `connection`, rather than mutually-exclusive primitive parameters
* `check_rollout` no longer ignores `--do-not-contact-google-play`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [upcoming]
+
+### Changed
+* Separates read-only logic from logic that requires a transaction
+
 ## [4.1.0] - 2019-07-10
 ### Removed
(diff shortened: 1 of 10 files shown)


mitchhentges:
After some discussion ending [here](https://github.com/mozilla-releng/mozapkpublisher/pull/213#issuecomment-514767189), we're not going to use types on the API boundary

PR #219, opened 2019-08-01
Amazon support

mitchhentges:
This still needs significant work (fix tests, add new tests, update other google play functions, fix/re-add docs), but I'd like to make sure that my base approach is :+1:-worthy.

I wouldn't recommend reviewing this PR, except for the conversation at `push_apk(...)` :smile:


mozapkpublisher/push_apk.py
```diff
@@ -1,59 +1,33 @@
...
 from mozapkpublisher.common.exceptions import WrongArgumentGiven
+from mozapkpublisher.common.store import AmazonStoreEdit, GooglePlayEdit
 
 logger = logging.getLogger(__name__)
 
 
+ExtractedApk = namedtuple('ProcessedApk', ['file', 'metadata'])
+
+
 def push_apk(
```
mitchhentges: @JohanLorenzo I'd like you to take a look at `push_apk(...)` here.

When called by a consumer like `pushapkscript`, usage will look something like this:
```
def pushapkscript_function(publish_config):
    apks = ['file', 'file2'] # would come from external source
    expected_package_names = ['org.mozilla.fenix'] # would come from external source

    def upload_google(package_name, extracted_apks):
        with open(publish_config['google_credentials_file']) as credentials:
            with GooglePlayEdit.transaction(
                publish_config['google_service_account'],
                credentials,
                package_name,
                contact_server=publish_config['contact_server'],
                commit=publish_config['commit']
            ) as edit:
                edit.update_app(extracted_apks, publish_config['track'],
                                publish_config['rollout_percentage'])

    def upload_amazon(package_name, extracted_apks):
        with AmazonStoreEdit.transaction(
            publish_config['amazon_client_id'],
            publish_config['amazon_client_secret'],
            package_name,
            contact_server=publish_config['contact_server'],
            commit=publish_config['commit']
        ) as edit:
            edit.update_app(extracted_apks)

    push_apk(
        apks,
        upload_amazon if publish_config['target_platform'] == 'amazon' else upload_google,
        expected_package_names,
        skip_checks_fennec=True,
    )
```

Note that some `if/else` logic is happening in `pushapkscript` here, which is one of the points you mentioned in previous `mozapkpublisher` discussion.

However, I'm not sure the existing approach scales well to include amazon support.
Here's what I got when I tried massaging the existing approach:
```
# in pushapkscript
def pushapkscript_function(publish_config):
    apks = ['file', 'file2'] # would come from external source
    expected_package_names = ['org.mozilla.fenix'] # would come from external source

    google_credentials_file = None
    if publish_config['target_platform'] == 'google':
        google_credentials_file = open(publish_config['google_credentials_file'])

    push_apk(
        apks,
        expected_package_names,
        publish_config['target_platform'],
        publish_config.get('amazon_client_id'),
        publish_config.get('amazon_client_secret'),
        publish_config.get('google_service_account'),
        google_credentials_file,
        publish_config.get('google_track'),
        publish_config.get('google_rollout_percentage'),
        commit=publish_config['commit'],
        contact_server=publish_config['contact_server'],
        skip_checks_fennec=True,
    )

    if publish_config['target_platform'] == 'google':
        google_credentials_file.close()

# ------
# here in mozapkpublisher
def push_apk(
    apks,
    expected_package_names,
    target_platform,
    amazon_client_id=None,
    amazon_client_secret=None,
    google_service_account=None,
    google_credentials_file=None,
    google_track=None,
    google_rollout_percentage=None,
    commit=True,
    contact_server=True,
    skip_check_ordered_version_codes=False,
    skip_check_multiple_locales=False,
    skip_check_same_locales=False,
    skip_checks_fennec=False,
):
    if target_platform == "google" and (
        google_service_account is None
        or google_credentials_file is None
        or google_track is None
    ):
        raise ValueError('When "target_platform" is "google", the account, credentials and track '
                         'must be provided')

    if target_platform == "amazon" and (
        amazon_client_id is None
        or amazon_client_secret is None
    ):
        raise ValueError('When "target_platform" is "amazon", the client_id and client_secret '
                         'must be provided')

    # We want to tune down some logs, even when push_apk() isn't called from the command line
    main_logging.init()

    apks_metadata_per_paths = extract_and_check_apks_metadata(
        apks,
        expected_package_names,
        skip_checks_fennec,
        skip_check_multiple_locales,
        skip_check_same_locales,
        skip_check_ordered_version_codes,
    )

    # Each distinct product must be uploaded in different "edit"/transaction, so we split them
    # by package name here.
    apks_by_package_name = _apks_by_package_name(apks_metadata_per_paths)
    for package_name, extracted_apks in apks_by_package_name.values():
        if target_platform == 'amazon':
            with AmazonStoreEdit.transaction(amazon_client_id, amazon_client_secret, package_name,
                                             contact_server=contact_server, commit=commit) as edit:
                edit.update_app(extracted_apks)
        elif target_platform == 'google':
            with store.edit(google_service_account, google_credentials_file.name,
                            package_name, contact_google_play=contact_server,
                            commit=commit) as edit:
                edit.update_app(extracted_apks, google_track, google_rollout_percentage)
        else:
            raise ValueError(f'Unexpected target platform of "{target_platform}", expected either'
                             f'"amazon" or "google"')
```
Notes:
* Managing `google_credentials_file` in `pushapkscript` isn't ergonomic
* What used to be a `publish_config['$platform_specific_option']` call in the callback approach (that forced required values to exist, otherwise it would error) has become much more verbose and manual. As parameters are added/removed to the function, maintainers have to remember to manually check 'em here
* Related to the last point: as a consumer of `push_apk`, it's harder to tell what parameters are required, and which can be `None`. With the callback approach, you know which parameters are required because you're using them in the callback body
* FWIW, setting up the callbacks is a little tough to understand, but I believe that on the whole, `push_apk(...)` with the callback is easier to understand than all the parameters. For example, it's easier to forget to close `google_credentials_file`, but the callbacks approach let's you use techniques (like context managers) to reduce that risk.

Johan, what do you think? Is there an approach that I'm missing that will be a stronger choice than the callback approach?

JohanLorenzo: Replied in https://github.com/mozilla-releng/mozapkpublisher/pull/219#pullrequestreview-269516165

mitchhentges:
@JohanLorenzo 

> I can easily see the data it was given and notice if something is wrong. 

FWIW, you can pdb at the callsite (rather than within `push_apk(...)`) and you'll see the same data.

> So, let's just change push_apk() to take a file path instead of a file handler.

Yeah, in this case, perhaps it's ok to push IO down further into the library. The call site (`ServiceAccountCredentials.from_p12_keyfile(...)`) expects a file name, not a file, so there's no reason to open the file in advance :+1:

> In the [callback] case, you have to update 4 different places, including 2 which live in a different module.

I'm not too worried by hard-to-get-wrong migrations like adding parameters, but FWIW, I'm not sure that the callback approach requires more parameter changes. e.g.: if `update_app` changes (e.g. Amazon allows a "targeting" parameter) , you have to update the following places:

###### existing solution:
* `update_app` params and call to `update_app` in `push_apk`
* `push_apk` params and external calls (`pushapkscript`)

###### callback solution:
* `update_app` params and external calls (`pushapkscript`)

> In the callback method, the push_apk() signature just tells you a function is expected, not what its parameters are

It's true that you're not plugging parameters in (you have to know the existence of `GooglePlayEdit` and `AmazonStoreEdit`) which isn't ideal. However, when you're using `push_apk(...)` for the first time, it's obvious that you need to fill the callback with _something_, and you're pushed to learn what goes inside. Once you find the `Edit` classes, it's hard to provide the wrong parameters.
However, with the existing approach, it's easier to fail at runtime/production with code that "looks" like it's provided all the necessary parameters.

(FWIW, I made a dummy [gist](https://gist.github.com/mitchhentges/bf38f085d1566cf0373970b01fc0e956) that keeps `push_apk(...)` minimal (all it does is ensure checks pass before running the push), while keeping calls easy to understand (the dynamic callbacks are replaced by implementations for each target store). Additionally, it keeps `push_apk` as the "sole interface", which was one of the goals you mentioned. It's a little boiler-plate-y without `@dataclass`, though, if that's a concern)

-----

My core concern with the existing solution is that it's representing multiple disparate workflows in a single generic function. Maintainable code isn't code that requires the smallest amount of lines to change: it's code that can be updated without being surprised by unexpected behaviour/usages.
In the example you provided:

* `secret` means entirely different things depending on context. It's either the actual password itself, or a _filename_.
* As someone using the function, I need to make the connection that `user_name` means both "service account" and "client id". Like `secret`, there's two different concepts being blended together
* The `kwarg` work isn't the most obvious, and will become more unwieldy as more store-specific parameters appear

Not that I like to dogmatically follow "programming principles", but if the above concerns aren't sufficiently concrete, it could be worth considering Uncle Bob's [Single Responsibility Principle](https://en.wikipedia.org/wiki/Single_responsibility_principle)

-----

Thank you for investigating the approach and explaining your thoughts - with you going away next Tuesday, I'm glad that I have a clear picture of what to do to land this soon. Though I've got some concerns, I'm going to :man_shrugging: again and implement the solution you described.

mozapkpublisher/common/apk/extractor.py
```diff
@@ -42,8 +43,11 @@ def extract_metadata(original_apk_path, extract_firefox_metadata):
         metadata['version_code'] = androguard_apk.get_androidversion_code()
 
         with ZipFile(apk_copy.name) as apk_zip:
-            metadata['architecture'] = _extract_architecture(apk_zip, original_apk_path)
-            metadata['locales'] = _extract_locales(apk_zip)
+            if extract_architecture_metadata:
+                metadata['architecture'] = _extract_architecture(apk_zip, original_apk_path)
```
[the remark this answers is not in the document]
mitchhentges: Yeah, we'll only avoid extracting architecture metadata if we `skip_check_ordered_version_codes`. This was useful for publishing a dummy app to the Amazon Store

mozapkpublisher/common/store.py
```diff
@@ -51,6 +54,129 @@ def execute(self):
         return self._return_value
 
 
+def http(expected_status, method, url, **kwargs):
```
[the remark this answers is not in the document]
mitchhentges: I'm not a big fan of throwing retries at network faults, since if Amazon is really down, it's just delaying us from being aware of the issue (since we have to wait for all the additional retries to go through). If we have network issues, shouldn't we address those directly?

However, we use `retry` elsewhere already, so I'll include this suggestion.
Note: we don't retry requests to Google, though :thinking:

mozapkpublisher/push_apk.py
```diff
@@ -93,16 +61,14 @@ def _get_ordered_version_codes(apks):
...
-    parser.add_argument(
-        '--track',
-        action='store',
-        required=True,
-        help='Track on which to upload'
-    )
-    parser.add_argument(
+    google_parser = subparsers.add_parser('google')
+    google_parser.add_argument('track', help='Track on which to upload')
+    google_parser.add_argument('--service-account', help='The service account email', required=True)
```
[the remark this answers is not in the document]
mitchhentges: :grimacing: I'm not sure the benefit of hiding the platform-specific jargon behind a help message - **especially** when the "secret"-y parameter either is a string or is a reference to a file.
I'll match your request since you're on PTO, though

mozapkpublisher/common/store.py
```diff
@@ -101,14 +235,20 @@ def upload_apk(self, apk_path):
                     return
             raise
 
-    def update_track(self, track, version_codes, rollout_percentage=None):
+    def update_track(self, track, version_codes, rollout_percentage):
```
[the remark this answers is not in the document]
mitchhentges: Good point! I'll condense it into `update_app()`

mitchhentges:
@MihaiTabara this is ready for full review (though I want to update `pushapkscript` and work on Firefox TV automation before landing this

mozapkpublisher/common/store.py
```diff
@@ -144,22 +282,22 @@ def update_whats_new(self, language, apk_version_code, whats_new):
...
-        logger.info('Changes committed')
-        logger.debug('edit_id "{}" for "{}" has been committed'.format(edit_id, package_name))
-    else:
-        logger.warning('Transaction not committed, since `commit` was `False`')
-
-
-def edit_resource_for_options(contact_google_play, service_account, credentials_file_name):
+    @staticmethod
+    @contextmanager
+    def transaction(service_account, credentials_file_name, package_name, *, contact_server, commit):
```
MihaiTabara: +1 for consistent naming of `contact_server`

mozapkpublisher/common/store.py
```diff
@@ -180,7 +318,7 @@ def edit_resource_for_options(contact_google_play, service_account, credentials_
 
         return service.edits()
     else:
-        logger.warning('Not a single request to Google Play will be made, since `contact_google_play` was set')
+        logger.warning('Not a single request to Google Play will be made, since `contact_google_play` was set to `False`')
         edit_resource_mock = MagicMock()
```
MihaiTabara: Not blocking this PR: I really liked the defined `MockAmazonStoreEdit` from https://github.com/mozilla-releng/mozapkpublisher/pull/219/files#diff-0eca0f1f55bbb0ccdc047ed838061cc0R164. Slightly more straightforward to have it defined in the first place, rather than doing the processing here. Maybe we should revisit this part in the future. 🤔

mozapkpublisher/common/store.py
MihaiTabara: Later edit: Ah, you need them consistent for rewriting `push_apk()` in https://github.com/mozilla-releng/mozapkpublisher/pull/219/files#diff-45a657627129bc0d2b7ae44372d28670R51. Neat 👍

mozapkpublisher/push_apk.py
```diff
@@ -3,56 +3,79 @@
...
     """
+    if target_store == "google" and track is None:
+        # The Google store allows multiple stability "tracks" to exist for a single app, so it
+        # requires you to disambiguate which track you'd like to publish to.
+        raise ValueError('When "target_store" is "google", the track must be provided')
+    if target_store == "amazon":
+        # The Amazon app doesn't have a stability "tracks" tool like Google. It _does_ have a
+        # "Live App Testing" mechanism, but you have to use the website to use it (the API
+        # doesn't support it). So, it's always the "production" app that's updated, and there's
+        # no need to specify "track"
```
MihaiTabara: Given that the Amazon API is young, I still hope this will change in the future.

mozapkpublisher/push_apk.py
```diff
@@ -3,56 +3,79 @@
...
+        # Source:  "Play Store API supports additional resources (such as testers and tracks) that
+        # the [Amazon] Appstore currently does not support."
+        # https://developer.amazon.com/docs/app-submission-api/migrate.html
+        if track is not None:
+            raise ValueError('Tracks are not supported on Amazon')
```
MihaiTabara: Question: any reason we're no longer raising `WrongArgumentGiven` but `ValueError` here and two lines below?

mozapkpublisher/push_apk.py
```diff
@@ -111,18 +131,40 @@ def main():
         help='The percentage of user who will get the update. Specify only if track is rollout'
     )
 
+    subparsers.add_parser('amazon')
```
MihaiTabara: Nit: are we actually using this anywhere? I don't see we grab the object like we do in line `123`.

mozapkpublisher/test/common/test_store.py
```diff
@@ -0,0 +1,394 @@
...
+
+
+def test_google_update_whats_new(edit_resource_mock):
+    google_play = GooglePlayEdit(edit_resource_mock, 1, 'dummy_package_name')
+
+    google_play.update_whats_new('en-GB', '2015012345', 'Check out this cool feature!')
+    edit_resource_mock.apklistings().update.assert_called_once_with(
+        editId=google_play._edit_id,
+        packageName='dummy_package_name',
+        language='en-GB',
```
MihaiTabara: +1 for the locale chosen :P

mozapkpublisher/common/store.py
```diff
@@ -1,4 +1,4 @@
-""" googleplay.py
+""" store.py
 
     The way to get the API access is to
```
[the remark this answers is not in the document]
JohanLorenzo: Good catch! I wonder if we should just get rid of this string. Here's my reasoning: 

a. Releng has never generated any creds, we've relied on Relman.
b. Relman has another document where they explain the steps to get this cred generated. I've seen both Sylvestre and Julien use it. I've never seen them use `googleplay.py`.
c. These instructions haven't changed in (at least) 4 years https://github.com/mozilla-releng/mozapkpublisher/commit/83d6e6cf9ea898e51146752de1bfde2f10e76d41. 
d. The Google Play interface has changed several times over the past couple of years.

As a consequence, odds are these instructions are outdated. I believe we should get rid of them.

mozapkpublisher/common/store.py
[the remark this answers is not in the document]
JohanLorenzo: I'm okay to land this as is. Please note I believe the lack of a retry mechanism will bite us once this code goes to production.

> since if Amazon is really down, it's just delaying us from being aware of the issue (since we have to wait for all the additional retries to go through). If we have network issues, shouldn't we address those directly?

Network is not 100% reliable. We may hit errors while Amazon remains up and running. I agree it's delaying us from being aware of perma-failures. Although, it saves people from having to contact us and us from having to manually rerun tasks. We cannot realistically chase down all types of network errors. The retry mechanism is a trade-off that saves people's time except when a perma-failure happens. It's implemented widely across releng.

> we don't retry requests to Google, yet 🤔 

It's true we don't. Although, their library handles it for us: https://github.com/googleapis/google-api-python-client/blob/895205850f2825a49138d1240e882a38afe93ffa/googleapiclient/http.py#L849. 

> Perhaps we should wait until mozapkpublisher has been refactored to be async-friendly?

I'm sorry for the confusion. `mozpkapublisher` doesn't need to be async to have a retry mechanism. I just gave the scriptworker link as an example. As another example, Google has implemented a synchronous version https://github.com/googleapis/google-api-python-client/blob/895205850f2825a49138d1240e882a38afe93ffa/googleapiclient/http.py#L131

> I remember cases where the upload to Google Play went halfway through and we had to make manual changes in the Google Play store combined with manual rerun (e.g. one of the apks was successful but the other didn't and when rerunning it would reject based on the already existing version code)

My apologies, I don't remember such cases. I'm surprised it ever has happened. Google Play implements a commit mechanism. mozapkpublisher commits after all APKs are uploaded. I might be remembering wrong. Would you have a bug number?

mozapkpublisher/push_apk.py
JohanLorenzo: Good catch! :+1:

mozapkpublisher/common/store.py
```diff
@@ -51,6 +51,130 @@ def execute(self):
...
+        try:
+            yield edit
+            if commit:
+                edit.commit()
+                logger.info('Changes committed')
+            else:
+                logger.warning('Edit not committed, since `commit` was `False`. Validating and '
+                               'cancelling the edit...')
+                edit.validate()
+                edit.cancel()
```
JohanLorenzo: 👍

mozapkpublisher/push_apk.py
JohanLorenzo: This value is stored in `target_store` which is passed down to `push_apk()`

mozapkpublisher/test/common/test_store.py
```diff
@@ -0,0 +1,394 @@
...
+                                               'applications/dummy_package_name/edits/edit_id/'
+                                               'validate',
+                                  auth=ANY)
+        mock_http.assert_any_call(204, 'delete', 'https://developer.amazon.com/api/appstore/v1/'
+                                                 'applications/dummy_package_name/edits/edit_id',
+                                  headers={'If-Match': 'edit_etag'},
+                                  auth=ANY)
+
+
+def test_amazon_transaction_cancel_on_exception():
```
JohanLorenzo: Great test!

mozapkpublisher/common/store.py
mitchhentges: Yeah, Johan and I briefly discussed this [in slack here](https://mozilla.slack.com/archives/CDKB7QPCL/p1567092097016800). By adding a new layer to represent the raw communication to Google, we could remove this need for this magic mock that's used in production(!)

mozapkpublisher/push_apk.py
mitchhentges: Hmm, I wonder why we defined our own `WrongArgumentGiven`, when `ValueError` fills a very similar role. I see that it wraps `LoggedError`, which raises the question of whether unexpected exceptions (that don't sublass `LoggedError`) get logged?

mozapkpublisher/common/store.py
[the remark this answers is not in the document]
mitchhentges: Let's work on a "retry" patch after landing this - I'm guessing there'll be other issues as well (such as handling un-cleaned-up edits) that we'll want to debug and resolve as well

README.md
```diff
@@ -60,4 +60,4 @@ source venv/bin/activate
 ```
 
   * Note `beta` track on Google Play, that's our way to show to people on Play Store that it's not a finished product. We don't use the "production" track for Nightly, unlike beta and release.
-1. If all goes well, add `--commit` to the command line and rerun it.
+1. If all goes well, add `--commit` (or `--keep` for Amazon) to the command line and rerun it.
```
mitchhentges: I used a different CLI argument for this since:

&nbsp;|discard|keep|commit
---|---|---|---
Google|✔|✖ "edits" from automation aren't visible on the web interface, so there's no value in keeping one|✔
Amazon|✔|✔ handoff from automation to humans|✖ no use case yet

mozapkpublisher/common/store.py
```diff
@@ -26,13 +11,22 @@
 from unittest.mock import MagicMock
 
 from mozapkpublisher.common.exceptions import WrongArgumentGiven
+import requests
 
 logger = logging.getLogger(__name__)
 
 
+BASE_AMAZON_URL = 'https://developer.amazon.com/api/appstore/v1'
+AMAZON_RECENT_CHANGES = {
```
mitchhentges: These are pretty generic, and there's only one amazon app, so I figured it's worth hard-coding these in `mozapkpublisher` rather than taking these as arguments

JohanLorenzo: Okay, thanks for the context!

mozapkpublisher/common/store.py
```diff
@@ -26,13 +11,22 @@
...
+    'en-US': 'Bug fixes and technical improvements.',
+    'de-DE': 'Fehlerkorrekturen und Technische Verbesserungen.',
```
JohanLorenzo: Question: Has this string been validated by the l10n team?

mozapkpublisher/common/store.py
```diff
@@ -26,13 +11,22 @@
...
+    'fr-FR': 'Correction de bugs et amélioration des techniques.',
```
JohanLorenzo: Nit: Same question here. As an `fr-FR` native speaker, I'd say: 

> Corrections de bugs et améliorations techniques.

instead.

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [6.0.0] - TBD
```
JohanLorenzo: Nit: This needs a small rebase. I shipped 6.0.0 today.

mozapkpublisher/common/store.py
mitchhentges: These are the strings that FFTV has been using before.
Since we aren't adding/changing the strings, I don't think we need to have them validated by l10n

mozapkpublisher/common/store.py
mitchhentges: That makes sense :+1: but if I change this string, do I need to talk to l10n then?

JohanLorenzo: Let's keep it as is, then :)

PR #221, opened 2019-08-14
Fixes #220, allows any track to be "rolled out" to

### LINKED ISSUE #220: partial rollout shouldn't set status to completed
I believe this is a regression from moving to the google play publishing API v3.  Instead of using a "rollout" track, partial rollouts can now be done on any track (typically "production"), but the status needs to be set appropriately.

https://developers.google.com/android-publisher/tracks#staged_rollouts
https://developers.google.com/android-publisher/api-ref/edits/tracks#resource

mitchhentges:
I'm going to test this by [rerunning this task](https://taskcluster-artifacts.net/QlgP9oEBR96lKkt0YeKZlg/0/public/logs/live_backing.log) before we land this.

Ready for CR, though :)


mitchhentges:
I've resolved the issue with rollout not working - in fact, none of the changes were being committed. Unfortunately, since we didn't want to affect too many `reference-browser` users when testing, we were only using `rollout`. Additionally, this coupled with the original rollout-fixing code for #220 being tested at the same time as the large transaction refactor (#213), meant that "commits not being applied" bug from #213 was incorrectly conflated.

Anyways, [this commit](https://github.com/mozilla-releng/mozapkpublisher/pull/221/commits/f83983a8919c854af40ed2d9e6a8e9daeb218e6b) resolves the issue.

PR #207, opened 2019-06-24
Bug 1560713 - Move to Google API v3

JohanLorenzo:
Fixed. It was a bug on the Google Play server. ~~Please don't review yet, the code is currently buggy. It returns:~~ 
```
2019-06-24 17:21:17,203 - extractor.py - INFO - Extracting metadata from a copy of "tmp/target-1.apk"...
2019-06-24 17:21:17,516 - extractor.py - INFO - Extracting metadata from a copy of "tmp/target-3.apk"...
2019-06-24 17:21:17,788 - extractor.py - INFO - Extracting metadata from a copy of "tmp/target-4.apk"...
2019-06-24 17:21:18,012 - extractor.py - INFO - Extracting metadata from a copy of "tmp/target.apk"...
2019-06-24 17:21:18,220 - checker.py - INFO - Checking APKs' metadata and content...
2019-06-24 17:21:18,221 - checker.py - INFO - Expected product types ['org.mozilla.fennec_aurora'] found
2019-06-24 17:21:18,221 - checker.py - INFO - Firefox version "68.0a1" matches package name "org.mozilla.fennec_aurora"
2019-06-24 17:21:18,221 - checker.py - INFO - All APKs have the same Firefox version: 68.0a1
2019-06-24 17:21:18,221 - checker.py - INFO - All APKs have the same Firefox BuildID: 20190621074959
2019-06-24 17:21:18,221 - history.py - DEBUG - Expected to find these combos for Firefox 68.0a1: x86 API 16+, arm64-v8a API 21+, x86_64 API 21+, armeabi-v7a API 16+
2019-06-24 17:21:18,221 - checker.py - INFO - Every expected APK was found!
2019-06-24 17:21:18,221 - checker.py - INFO - "tmp/target-1.apk" is multilocale.
2019-06-24 17:21:18,221 - checker.py - INFO - "tmp/target-3.apk" is multilocale.
2019-06-24 17:21:18,221 - checker.py - INFO - "tmp/target-4.apk" is multilocale.
2019-06-24 17:21:18,221 - checker.py - INFO - "tmp/target.apk" is multilocale.
[...]

PR #223, opened 2019-08-16
Revert to android-publisher API v2

### LINKED ISSUE #222: Downgrade to Google API v2 to support rollout again
Until #220 is resolved, we should downgrade to the Google `android-publisher` API v2 (instead of v3, which is causing issues for us)

mitchhentges:
Fixes #222.

Reverts the API usage changes from [this PR](https://github.com/mozilla-releng/mozapkpublisher/pull/207).


mitchhentges:
This is only needed if we're blocked on a Fennec release and need `rollout` to work.
Otherwise, let's wait and see if #220 can be fixed first

mitchhentges:
#221 was landed, this isn't needed anymore

PR #257, opened 2024-01-15
fix(deps): Pin androguard<4

hneiva:
Androguard version 4 moved some packages around and causes import exceptions. There might be other changes that will required further testing before using androguard 4.x


jcristau (approved):
Let's open an issue about androguard 4.x compat, please.

PR #260, opened 2024-06-27
Bug 1905093 - set googleapiclient retries to 3

gbrownmozilla:
Confirmed num_retries at https://googleapis.github.io/google-api-python-client/docs/epy/index.html


jcristau (changes requested):
I don't think this is enough?  AFAICT `googleapiclient.discovery.build` just uses num_retries for the `_retrieve_discovery_doc` request, but AFAICT we also need it everywhere else we make requests (i.e. pretty much all calls to `execute` in store.py)?

gbrownmozilla:
You are probably right. I haven't had a chance to look into this today; feel free to take it, or I can have another look after pto.

jcristau (approved):

PR #265, opened 2025-03-24
Remove get_apk.py

Eijebong:
It's was broken 6 years ago by 1d4775dffde4a82ee3dca257f1f07d42c5a6f8c9 Since it's not used, is broken and references fennec artifacts which haven't been a thing since firefox android migrated to m-c, just remove it


README.md
```diff
@@ -44,7 +44,7 @@ source venv/bin/activate
 
 1. Generate a Google Play Store json certificate. This certificate needs to have write access to the app you want to publish. In this context, "app" means Fennec, Fennec Beta or Fennec Nightly.
 1. Execute the steps defined in the section above.
-1. Download the latest signed builds. For instance, for Fennec Nightly: `./mozapkpublisher/get_apk.py --latest-nightly`
+1. Download the latest [signed builds](`https://treeherder.mozilla.org/jobs?repo=mozilla-central&searchStr=signing-apk-fenix-nightly)
```
jcristau, line 47: Probably should be signing-bundle rather than signing-apk?

jcristau (approved):

README.md
Eijebong: You're right, my use case needs APKs so I took that URL from there but the line below is google specific. I changed it.

PR #268, opened 2025-03-25
Add a pre-commit hook to ensure the uv.lock file doesn't get out of sync

Eijebong:


Eijebong:
Marking as draft until we can get the pre-commit.ci app installed on the repo

PR #280, opened 2025-07-01
Bug 1974870 - Upload APKs to SGS with a unique name

Eijebong:
Turns out the API asks for a binarySeq for absolutely **no reason whatsoever** since they also expect every file name to be unique. I can't wait to discover that the filename cannot contain any `.` except for the extension or that there's a max size at 20 characters for no good reason. Given the rest of the API that would not even surprise me...

This is untested because we have to test in prod, they don't offer any way of having a test app; but that's how forms work, so hopefully they don't do anything stupid on their side...


jcristau (approved):
I guess we're assuming we won't re-use a version_name?  That seems less robust than using the version code, but maybe good enough as long as we don't e.g. ship nightly (or rc?) builds this way.

PR #325, opened 2026-06-18
feat!: add support to publish to huawei store

abhishekmadan30:


mozapkpublisher/huawei_api/__init__.py
```diff
@@ -0,0 +1,229 @@
...
+
+    async def upload_apks(self, package_name, apks, rollout_rate, submit=False):
+        """
+        Upload the APKs passed as arguments. The app to be updated will be inferred from
+        the package name.
+
+        If `rollout_rate` is not None, the submission uses `releaseType=2` (phased
+        release). Setting the target rollout percentage is not handled here; that's
+        managed in the AppGallery Connect console (or via a follow-up phased-release
+        endpoint that is not yet wired up).
```
JohanLorenzo, line 45: Do we intend to wire it?

mozapkpublisher/huawei_api/__init__.py
```diff
@@ -0,0 +1,229 @@
...
+
+    async def infer_app_id_from_package_name(self, package_name):
+        """
+        Returns the app ID related to the package name provided.
+        """
+        result = await self.api.app_id_list(package_name=package_name)
+        apps = result.get("appids", []) or []
+        for app in apps:
+            if app.get("package") == package_name:
+                return app["value"]
```
JohanLorenzo: Can we safeguard against the case where multiple apps match the same package name?

mozapkpublisher/huawei_api/__init__.py
```diff
@@ -0,0 +1,229 @@
...
+            logger.warning('No APKs were uploaded since `dry_run` was `True`')
+            return
+
+        app_id = await self.infer_app_id_from_package_name(package_name)
+
+        files = []
+        for apk in apks:
+            fd, metadata = apk
+
+            file_name = "{}-{}-{}.apk".format(metadata["package_name"], metadata["architecture"], metadata["version_name"])
```
JohanLorenzo: Nit: duplicated logic. Let's centralize it. 

https://github.com/mozilla-releng/mozapkpublisher/blob/1dd73f8157075719dca9dfe4116a533ad3680cb4/mozapkpublisher/sgs_api/__init__.py#L60

mozapkpublisher/huawei_api/__init__.py
```diff
@@ -0,0 +1,229 @@
...
+
+        await self.api.update_app_file_info(app_id, files)
+
+        if submit:
+            release_type = 2 if rollout_rate is not None else 1
+            await self.api.submit_app(app_id, release_type=release_type)
+
+    async def upload_file(self, app_id, file, name):
+        """
+        Uploads a file to the huawei app gallery and returns its `fileDestUrl`
```
JohanLorenzo: I'm not sure I understand what `fileDestUrl` is used for, afterwards. Would you mind explaining it to me?

mozapkpublisher/huawei_api/__init__.py
```diff
@@ -0,0 +1,229 @@
...
+    async def __aenter__(self) -> "HuaweiAppGalleryApi":
+        return self
+
+    async def __aexit__(self, *args: Any) -> None:
+        await self._client.close()
+
+    def _default_headers(self) -> Dict[str, str]:
+        """
+        Returns headers necessary for every authenticated request, according to
+        https://developer.huawei.com/consumer/en/doc/development/AppGallery-connect-Guides/agcapi-getstarted-0000001111845114
```
JohanLorenzo: Thank you for point it back to the original documentation!

mozapkpublisher/huawei_api/__init__.py
```diff
@@ -0,0 +1,229 @@
...
+        url = urljoin(base_url, route)
+
+        response = await self._client.request(method, url, headers=headers, **kwargs)
+
+        await raise_for_status_with_message(response)
+
+        body = await response.json()
+        if check_ret and isinstance(body, dict):
+            raise_for_ret_code(body)
+        return body
```
JohanLorenzo: Nit: Duplicated code. Let's extract this function in a central place 

https://github.com/mozilla-releng/mozapkpublisher/blob/1dd73f8157075719dca9dfe4116a533ad3680cb4/mozapkpublisher/sgs_api/__init__.py#L155-L171

mozapkpublisher/huawei_api/__init__.py
```diff
@@ -0,0 +1,229 @@
...
+
+    async def get_upload_url(self, app_id: str, suffix: str = "apk", release_type: int = 1) -> Dict[str, Any]:
+        """
+        Get an upload URL for a binary.
+
+        https://developer.huawei.com/consumer/en/doc/AppGallery-connect-References/agcapi-obtain-uploadurl-0000001158245059
```
JohanLorenzo: Link returns 404.

mozapkpublisher/huawei_api/__init__.py
```diff
@@ -0,0 +1,229 @@
...
+            "GET",
+            "/api/publish/v2/upload-url",
+            params={"appId": app_id, "suffix": suffix, "releaseType": release_type},
+        )
+
+    async def upload_file(self, upload_url: str, auth_code: str, file_path: str, name: str) -> Dict[str, Any]:
+        """
+        Upload a file using a previously obtained upload URL + auth code.
+
+        https://developer.huawei.com/consumer/en/doc/AppGallery-connect-References/agcapi-upload-file-0000001158245057
```
JohanLorenzo: Link returns 404 too.

mozapkpublisher/huawei_api/__init__.py
```diff
@@ -0,0 +1,229 @@
...
+        if not result_list:
+            raise HuaweiUploadException(
+                "The upload result didn't contain a fileInfoList entry: {}".format(body)
+            )
+        result = result_list[0]
+
+        if int(result.get("size", 0)) != original_file_size:
+            raise HuaweiUploadException(
+                "The upload result gave a file size different than what was uploaded. Got {}, expected {}".format(
+                    result.get("size"), original_file_size
```
JohanLorenzo, line 206: Good call on checking the file size. I see the SGS don't expose checksums. Does the Huawei Store do, by any chance? If not, let's also add a code comment.

https://github.com/mozilla-releng/mozapkpublisher/blob/1dd73f8157075719dca9dfe4116a533ad3680cb4/mozapkpublisher/sgs_api/__init__.py#L221-L227

mozapkpublisher/huawei_api/__init__.py
```diff
@@ -0,0 +1,229 @@
...
+                )
+            )
+
+        return result
+
+    async def update_app_file_info(self, app_id: str, files: List[Dict[str, str]]) -> Dict[str, Any]:
+        """
+        Update the file info bound to an app release.
+
+        https://developer.huawei.com/consumer/en/doc/AppGallery-connect-References/agcapi-update-file-info-0000001158365049
```
JohanLorenzo: 404

mozapkpublisher/huawei_api/__init__.py
```diff
@@ -0,0 +1,229 @@
...
+            fd, metadata = apk
+
+            file_name = "{}-{}-{}.apk".format(metadata["package_name"], metadata["architecture"], metadata["version_name"])
+            file_dest_url = await self.upload_file(app_id, fd.name, file_name)
+            files.append({"fileName": file_name, "fileDestUrl": file_dest_url})
+
+        await self.api.update_app_file_info(app_id, files)
+
+        if submit:
+            release_type = 2 if rollout_rate is not None else 1
```
JohanLorenzo: Nit: Let's extract 1 and 2 into named constants so it's easier to know what they mean.

mozapkpublisher/huawei_api/__init__.py
```diff
@@ -0,0 +1,229 @@
...
+            "GET", "/api/publish/v2/app-info", params={"appId": app_id}
+        )
+
+        return AppContentInfo(result.get("appInfo", {}))
+
+    async def app_id_list(self, package_name: str) -> Dict[str, Any]:
+        """
+        Return the appId list for a given package name.
+
+        https://developer.huawei.com/consumer/en/doc/AppGallery-connect-References/agcapi-appid-list-0000001158245055
```
JohanLorenzo: 404

mozapkpublisher/huawei_api/auth.py
```diff
@@ -0,0 +1,32 @@
...
+        async with session.post(
+            "https://connect-api.cloud.huawei.com/api/oauth2/v1/token",
+            headers=headers,
+            json=payload,
+        ) as resp:
+            await raise_for_status_with_message(resp)
+
+            result = await resp.json()
+
+            return cast(str, result["access_token"])
```
JohanLorenzo: Why is the cast necessary?

mozapkpublisher/huawei_api/content_info.py
```diff
@@ -0,0 +1,48 @@
...
+        return self._inner.get("releaseState")
+
+    def as_file_info_payload(self, files: List[Dict[str, str]]) -> Dict[str, Any]:
+        """
+        Build the body for `PUT /api/publish/v2/app-file-info`. `files` is a list of
+        `{"fileName": ..., "fileDestUrl": ...}` entries (`fileDestUrl` is returned by
+        the upload step). `fileType` is 5 for an Android RPK/APK/AAB binary.
+        """
+        return {
+            "fileType": 5,
```
JohanLorenzo: Nit: Let's extract it into a named constant.

mozapkpublisher/huawei_api/error.py
```diff
@@ -0,0 +1,23 @@
+class HuaweiException(Exception):
+    def __init__(self, message: str):
+        self.message = message
```
JohanLorenzo: Nit: duplicated logic 

https://github.com/mozilla-releng/mozapkpublisher/blob/1dd73f8157075719dca9dfe4116a533ad3680cb4/mozapkpublisher/sgs_api/error.py#L1-L3

mozapkpublisher/huawei_api/utils.py
```diff
@@ -0,0 +1,59 @@
...
+    elif resp.status == 403:
+        raise HuaweiAuthorizationException(error_message)
+
+    raise aiohttp.ClientResponseError(
+        resp.request_info,
+        resp.history,
+        status=resp.status,
+        message=error_message,
+        headers=resp.headers,
+    )
```
JohanLorenzo, line 33: Duplicated logic 

https://github.com/mozilla-releng/mozapkpublisher/blob/1dd73f8157075719dca9dfe4116a533ad3680cb4/mozapkpublisher/sgs_api/utils.py#L6-L41

mozapkpublisher/huawei_api/utils.py
```diff
@@ -0,0 +1,59 @@
...
+    Huawei wraps responses in a `ret` envelope (`{ret: {code: int, msg: str}, ...}`).
+    A non-zero `code` indicates an in-band failure even when the HTTP status is 200.
+    """
+    ret = body.get("ret")
+    if not ret:
+        return
+    code = ret.get("code")
+    if code in (None, 0):
+        return
+    raise HuaweiException("Huawei API returned ret.code={}: {}".format(code, ret.get("msg")))
```
JohanLorenzo: Is there a list of documented ret codes in case we stumble upon one in the future?

mozapkpublisher/test/huawei/test_auth.py
```diff
@@ -0,0 +1,60 @@
...
+)
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "status,response,expectation",
+    (
+        pytest.param(
+            200,
+            {"access_token": "jambonBeurre", "expires_in": 172800},
```
JohanLorenzo: Hahahaha, I didn't know SGS test suite was using this value. It's a type of sandwich in France 😋

mozapkpublisher/test/huawei/test_rollout.py
```diff
@@ -0,0 +1,173 @@
...
+        "package_name": "org.mozilla.firefox",
+        "architecture": architecture,
+        "version_name": version_name,
+        "version_code": 1000,
+        "api_level": 21,
+    }
+
+
+@pytest.mark.asyncio
+async def test_upload_apks_release_type_full(responses_mock):
```
JohanLorenzo: I'm having trouble to make sense of this test. What does it do?

mozapkpublisher/test/huawei/test_rollout.py
```diff
@@ -0,0 +1,173 @@
...
+
+        async with HuaweiAppGallery("client_id", "access_token") as huawei:
+            await huawei.upload_apks("org.mozilla.firefox", [(_FD(path), _apk_metadata())], None, submit=True)
+    finally:
+        os.unlink(path)
+
+
+@pytest.mark.asyncio
+async def test_upload_apks_release_type_phased(responses_mock):
+    """When a rollout rate is set and submit=True, releaseType=2 is used."""
```
JohanLorenzo: If this test and the one before just test the difference of behavior between a phased rollout or not, isn't there a simpler way of checking this?

mozapkpublisher/test/huawei/test_rollout.py
```diff
@@ -0,0 +1,173 @@
...
+            def __init__(self, name): self.name = name
+
+        async with HuaweiAppGallery("client_id", "access_token") as huawei:
+            await huawei.upload_apks("org.mozilla.firefox", [(_FD(path), _apk_metadata())], 10, submit=True)
+    finally:
+        os.unlink(path)
+
+
+@pytest.mark.asyncio
+async def test_upload_apks_uses_unique_filename(responses_mock):
```
JohanLorenzo: Same comment here. Isn't there a simpler way of testing the uniqueness of filenames?

mozapkpublisher/test/huawei/test_submit.py
```diff
@@ -0,0 +1,43 @@
...
+from .common import basic_auth_headers
+from mozapkpublisher.huawei_api.error import HuaweiAuthenticationException
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "release_type",
+    (1, 2),
+)
+@pytest.mark.parametrize(
```
JohanLorenzo, line 13: TIL we can chain `@pytest.mark.parametrize` decorator! 👍

mozapkpublisher/test/huawei/test_upload.py
```diff
@@ -0,0 +1,159 @@
...
+async def test_upload_file(huawei, responses_mock, status, response, expectation):
+    responses_mock.post(
+        "https://upload.example/foo",
+        status=status,
+        payload=response,
+    )
+
+    with tempfile.NamedTemporaryFile("w") as tmp, expectation as exc:
+        tmp.write("1" * 10)
+        tmp.flush()
```
JohanLorenzo: I see we are generating fake APK files several times. Should we extract this into a fixture?

mozapkpublisher/get_huawei_token.py
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env python3
+
+import argparse
+import asyncio
+
+from mozapkpublisher.huawei_api.auth import create_access_token
+
+
+async def main(args: argparse.Namespace) -> None:
+    print(await create_access_token(args.client_id, args.client_secret))
```
JohanLorenzo: I understand the SGS has a very peculiar way of handling authentication. In the case of Huawei, what is this script used for?

mozapkpublisher/common/utils.py
```diff
@@ -25,11 +25,13 @@ def filter_out_identical_values(list_):
...
 
 def add_push_arguments(parser):
-    parser.add_argument('--store', help='Store on which to upload', choices=['google', 'samsung'], default="google")
+    parser.add_argument('--store', help='Store on which to upload', choices=['google', 'samsung', 'huawei'], default="google")
     parser.add_argument('--secret', help='File that contains google credentials (json). This is only required if the store is google.')
     parser.add_argument('--sgs-service-account-id', help='The service account ID for the samsung galaxy store. This is only required if the store is samsung')
     parser.add_argument('--sgs-access-token', help='The access token for the samsung galaxy store. This is only required if the store is samsung')
-    parser.add_argument('--submit', help='Submit the submission for review. This doesn\'t change anything unless the store is samsung', action='store_true')
+    parser.add_argument('--huawei-client-id', help='The client ID for the huawei app gallery. This is only required if the store is huawei')
+    parser.add_argument('--huawei-access-token', help='The access token for the huawei app gallery. This is only required if the store is huawei')
```
JohanLorenzo: Not a change for this PR. I wonder if we should change the way we define CLI arguments when it comes to authentication. The list gets longer every time we add a new provider.

mozapkpublisher/common/utils.py
```diff
@@ -25,11 +25,13 @@ def filter_out_identical_values(list_):
...
+    parser.add_argument('--submit', help='Submit the submission for review. This doesn\'t change anything unless the store is samsung or huawei', action='store_true')
```
JohanLorenzo: What does submit do in the case of Huawei? If we don't provide `--submit` what happens? How different is it from `--dry-run`?

JohanLorenzo (changes requested):
Thank you for taking the time to make mozapkpublisher support. The overall architecture makes sense to me and I believe this is the direction we want to go to. Thank you also for providing links to the Huawei documentation, that's super helpful. I wouldn't be surprised if the servers don't exactly follow the documentation. How did the end-to-end testing go?

Moreover, having a PR description with what you tested, what you found surprising during the implementation, how the overall implementation of Huawei Store differs from the Samsung Galaxy Store (SGS), would have been helpful for a first round of review as well as for future readers. Could you provide these points?

mozapkpublisher/common/utils.py
abhishekmadan30: I agree, I can file that as an issue

mozapkpublisher/huawei_api/__init__.py
abhishekmadan30: I added it to the doc string for future reference

mozapkpublisher/common/utils.py
abhishekmadan30: Essentially, theres 2 flag system, `--commit` and `--submit`, so `--commit` uploads the APK and sets it as a draft and `--submit` then publishes it. You cant `--submit` unless `--commit` has been done and it will just assumes its a dry run. The other way around is not true though you can `--commit` without `--submit` which will just leave a draft APK uploaded

mozapkpublisher/get_huawei_token.py
abhishekmadan30: This essentially prints out an access token as it expires every 48 hours so would just need setup a cron job to rotate the access token

mozapkpublisher/huawei_api/__init__.py
abhishekmadan30: Yes, I will do that after its linked up with Gecko for now the percentage can be handled in kind.yml

mozapkpublisher/huawei_api/__init__.py
JohanLorenzo: Follow-up: I see we can pass a `sha256` in the request to https://connect-api.cloud.huawei.com/api/publish/v2/upload-url/for-obs: 

> | Parameter | Mandatory | Type | Description | 
> |------------|-----------|-------|------------|
> | sha256 | No | String(64) | SHA-256 hash value of the app package content. If this parameter is passed, the integrity of an uploaded app package will be verified. If this parameter is not passed, verification is not performed. | 

[Source](https://developer.huawei.com/consumer/en/doc/AppGallery-connect-References/agcapi-upload-url-new-0000001111685200#section798917276209). Let's calculate the checksum locally and have the Huawei store check it for us.

mozapkpublisher/huawei_api/__init__.py
```diff
@@ -0,0 +1,250 @@
...
+import aiohttp
+import logging
+import os.path
+
+BASE_URL = "https://connect-api.cloud.huawei.com/"
+logger = logging.getLogger(__name__)
+
+# `releaseType` values accepted by the AppGallery publishing API.
+RELEASE_TYPE_FULL = 1
+RELEASE_TYPE_PHASED = 2
```
JohanLorenzo: Nit:

```suggestion
RELEASE_TYPE_FULL_ROLLOUT = 1
RELEASE_TYPE_PHASED_ROLLOUT = 2
```

so it's clearer what is full or phased.

mozapkpublisher/huawei_api/utils.py
JohanLorenzo: Nit: Thanks for adding https://developer.huawei.com/consumer/en/doc/AppGallery-connect-References/agcapi-publishingapi-errorcode-0000001163523297 to the comments. Could you also make this URL part of the exception message? This way, anyone who wants to debug an error can get to that webpage by just reading the logs.

mozapkpublisher/test/huawei/test_rollout.py
JohanLorenzo: Ctrl+F "phased" doesn't find any tests for phased rollouts. How do we test the `full/phased` behavior?

mozapkpublisher/test/huawei/test_rollout.py
JohanLorenzo: `test_upload_apks_uses_unique_filename` still mocks a few things. At a first glance, this seems unnecessary to test a simple function like `apk_file_name()`. What justifies the mock out? Also, could we add a verb in the function name (something like `get_` or `build_`). Function do things so conveying this information with a verb is usually a good practice.

mozapkpublisher/get_huawei_token.py
JohanLorenzo: Blocker: Oh, that's a very uncommon way of handling authentication. Could you explain how the authentication works in the current implementation? How would the cron job rotate the credentials stored in SOPS and have it deploy to the pushapk instances? Are there multiple ways of solving the authentication problem or are we forced to comply with that one? 

I'm flagging this comment as a blocker because we can't ship this PR without having clear ideas on how authentication has to work.

mozapkpublisher/common/utils.py
JohanLorenzo: WFM. Please do.

mozapkpublisher/common/utils.py
JohanLorenzo: Thanks for the explanation. SGTM.

JohanLorenzo (changes requested):
Thanks for handling almost all the comments I raised. This looks much cleaner. I'm marking it as "request changes" because what I understand of the authentication tells me this is likely not going work in production. Maybe there's something I misunderstood and I'd be happy to be convinced otherwise 🙂 More details about the authentication part below.

mozapkpublisher/get_huawei_token.py
abhishekmadan30: Ok I think I have solution. I was using the Client Service API (the client_credentials flow), and that is what forced the whole rotation design. After my chat with Bastien, I was seeing if the Huawei API had another option and I have switched to Huawei's Service Account mode, which removes the rotation problem entirely.

Now,the worker holds a static Service Account key (the private.json with key_id, sub_account, private_key) and signs a one hour PS256 JWT itself on each run. In this mode the JWT is the token, so there is no token exchange call, just Authorization: Bearer <jwt>. (See the [AppGallery Connect docs](https://developer.huawei.com/consumer/en/doc/appgallery-connect-guides/agc-remoteconfig-getapicredentials-0000002539183415).)

mozapkpublisher/test/huawei/test_rollout.py
```diff
@@ -0,0 +1,85 @@
...
+        responses_mock.post(
+            f"https://connect-api.cloud.huawei.com/api/publish/v2/app-submit?appId={APP_ID}&releaseType={submit_release_type}",
+            payload={"ret": {"code": 0, "msg": "ok"}},
+        )
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "rollout_rate,expected_release_type",
+    [(None, RELEASE_TYPE_FULL_ROLLOUT), (10, RELEASE_TYPE_PHASED_ROLLOUT)],
```
abhishekmadan30, line 65: @JohanLorenzo Sorry since I updated the PR, I am having issues submitting replies to your previous comments. I had keep the phase rollout, before by I had just paramterized it, into its number format but I added some statics to clear that up

JohanLorenzo (commented):
New authentication mechanism looks better to me. Thank you! I'm not approving yet because I want to know how the end-to-end testing went. The mocked out integration tests are good but not sufficient for a first-time implementation. Can we have some logs to show that it worked? More info in https://github.com/mozilla-releng/mozapkpublisher/pull/325#pullrequestreview-4531075492

PR #331, opened 2026-07-06
Bug 2052916 - Use the dedicated APIs to add/delete files to content IDs

Eijebong:
Starting July 2026, the samsung galaxy store stopped accepting content updates that modified the binary list, forcing users to instead use dedicated APIs for management.

They added a `binarySeqForDeviceInfo` parameter to the binary addition. After testing I found that it dropped support for 200+ devices on focus, and that it works fine without. So I didn't bother adding support for it since the documentation is unclear about what it even does (it probably needs to be keyed by architecture, but again, it works without it
anyway).

The one change in behavior I saw is that the UI now shows the filename as uploaded instead of some weird `focus-15206-armeabi-v7a.apk`. I couldn't figure out what was setting that name before, so I assume it's something Samsung did. The new name is the one we pass in the multipart form, including dots which the web UI doesn't allow you to upload manually.

It's not causing any issue, and we know they were storing that multipart name before because it caused bug 1974870. So nothing to do here, just noting it because this is a reminder that this name has to be unique if it's ever changed.

### DIFF
--- a/mozapkpublisher/sgs_api/__init__.py
+++ b/mozapkpublisher/sgs_api/__init__.py
@@ -51,46 +51,33 @@ class SamsungGalaxyStore:
         current_info = content_info[0]
 
         # Because infer_content_id_from_package_name relies on a binary with the package name existing, we know that binary_list here is not empty
-        last_binary = max((binary for binary in current_info.binary_list), key=lambda binary: int(binary["binarySeq"]))
-        last_binary_id = int(last_binary["binarySeq"])
+        existing_binaries = current_info.binary_list
+        last_binary = max(existing_binaries, key=lambda binary: int(binary["binarySeq"]))
 
+        gms = last_binary["gms"]
+
+        # Transition the app to `REGISTERING`
+        await self.api.update_content_info(current_info)
+
+        binaries_oldest_first = sorted(existing_binaries, key=lambda binary: int(binary["binarySeq"]))
+
+        binaries_to_delete = max(0, len(existing_binaries) + len(apks) - 20)
+        for binary in binaries_oldest_first[:binaries_to_delete]:
+            await self.api.delete_binary(content_id, binary["binarySeq"])
+
+        new_binary_seqs = []
         for apk in apks:
             fd, metadata = apk
 
             file_name = "{}-{}-{}.apk".format(metadata["package_name"], metadata["architecture"], metadata["version_name"])
             file_key = await self.upload_file(fd.name, file_name)
-            new_binary = {
-                "fileName": os.path.basename(fd.name),
-                "versionCode": metadata["version_code"],
-                "binarySeq": last_binary_id + 1,
-                "packageName": metadata["package_name"],
-                "apiminSdkVersion": metadata["api_level"],
-                "apimaxSdkversion": None,
-                "iapSdk": last_binary["iapSdk"],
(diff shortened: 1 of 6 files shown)


mozapkpublisher/sgs_api/__init__.py
```diff
@@ -267,6 +259,48 @@ async def update_content_info(
...
+        the `REGISTERING`/`UPDATING` state.
+
+        https://developer.samsung.com/galaxy-store/galaxy-store-developer-api/content-publish-api/delete-binary.html
+        """
+        data = {
+            "contentId": content_id,
+            "binarySeq": binary_seq,
+        }
+
+        return await self._request("DELETE", "/seller/v2/content/binary", json=data)
```
bhearsum: This needs testing; it's unclear based on https://developer.samsung.com/galaxy-store/galaxy-store-developer-api/content-publish-api/delete-binary.html whether the data is in the body or the query string. (It says body, but the example uses the query string.)

bhearsum (commented):
Just for posterity / cross linking, https://github.com/bhearsum/mozapkpublisher/pull/new/more-tests is what I came up with. It looks like we converged on very similar solutions.

Eijebong:
Asked relman to let me do the next upload manually, waiting on that to be able to test any of this.

I looked at the store and found no information that'd help answer any of the questions... The UI is as good as the last time I've touched this.

Eijebong:
- Is binary_seq_for_device_info useful? Is copying from the latest binary_seq the right thing to do? (especially concerned if architectures don't match then... We might have to copy from the latest binary that matches the current arch. Not even sure if arch is relevant or not for the filtering)

We do **not** want to use that. It works fine without, and with the one from the current patch, it drops support for 200+ devices.

- Do we still have a 20 binary limit for a given content ID? I couldn't find anything about this anymore
 
Yup, you can only see this within an app in update mode...
image

- Does sending back the exact same content for a content ID actually put it in an updating mode?

Yup

- We no longer send versionCode / packageName / apiminSdkVersion at all. The old binaryList entry carried those explicitly. Hopefully samsung derives those properly from our APKs.

Yup, looks correct. The one thing that seems to have changed is the file name displayed in the binary list. I cannot figure out where the old name comes from (`focus-15206-armeabi-v7a.apk`). It's now the one we send in the multipart form, including dots in the version, which you can't even do from the web UI... `org.mozilla.focus-arm64-v8a-153.0.apk`

We must keep in mind here that this name needs to be unique if we feel like changing it. I'll make sure to explicit that in the commit message

mozapkpublisher/sgs_api/__init__.py
Eijebong: Tested. The example was correct, the docs not so much. Changed the code to match

bhearsum (approved):
> We do not want to use that. It works fine without, and with the one from the current patch, it drops support for 200+ devices.

Not blocking this, but I think it would be useful to try setting `binary_seq_for_device_info` on a per-arch basis, or do something else to help us understand its purpose. (I'm a bit concerned that there may be some downsides we're not aware of to not setting it.) This can probably just be filed as a follow-up for now.

Eijebong:
Yeah, it didn't exist before and omitting it gives us the same behavior so I honestly dig much more.

PR #333, opened 2026-09-02
feat!: add support to publish to huawei store

hneiva:
Closes #325 

Adds `--store huawei` to `push_apk`, publishing to the Huawei AppGallery Connect Publishing API: resolves the app ID from the package name, uploads each APK, binds them to the draft version, and optionally submits for release (`--submit`), with support for phased rollouts via `--rollout-percentage`. Authentication uses a self-signed Service Account JWT as the bearer token. Shared request/error-handling helpers are extracted into `common/store_api.py` and reused by the existing Samsung client.

### Local / Manual testing done:
Verified against the production AppGallery API with a real Service Account. A 155.0 arm64-v8a APK was uploaded to `org.mozilla.firefox`; `package-list` confirms the stored binary matches the local file byte-for-byte (same SHA-256 and size), under the filename `build_apk_file_name` generates. This exercised app ID resolution, the upload-URL and upload steps, and `app-file-info` binding. Authentication was confirmed separately: the JWT is accepted directly with no token exchange and no `client_id` header. The upload was run without `--submit`.


mozapkpublisher/huawei_api/__init__.py
```diff
@@ -7,15 +8,93 @@
...
+    if not 0 < rollout_rate <= 100:
+        raise HuaweiUpdateException(
+            "Rollout percentage must be in (0, 100]. Value given: {}".format(rollout_rate)
+        )
+
+    start_time = start_time or datetime.now(timezone.utc)
+    percent = "{:.2f}".format(rollout_rate)
+    return {
+        "phasedReleaseStartTime": start_time.strftime(PHASED_ROLLOUT_TIME_FORMAT),
+        "phasedReleaseEndTime": (start_time + PHASED_ROLLOUT_WINDOW).strftime(PHASED_ROLLOUT_TIME_FORMAT),
```
JohanLorenzo, line 97: Nit: why not using [`datetime.isoformat()`](https://docs.python.org/3/library/datetime.html#datetime.datetime.isoformat)?

mozapkpublisher/huawei_api/__init__.py
```diff
@@ -7,15 +8,93 @@
...
+# later, which the 10 minute budget covers.
+# https://developer.huawei.com/consumer/en/doc/AppGallery-connect-Guides/agcapi-pub-releasenotes-0000001111685226
+SUBMIT_MINIMUM_RETRY_DELAY = 120
+SUBMIT_RETRY_DELAY = 120
+SUBMIT_RETRY_TIMEOUT = 600
+
+# `ret.code` values `app-submit` returns while a package is still being processed.
+# https://developer.huawei.com/consumer/en/doc/AppGallery-connect-References/agcapi-publishingapi-errorcode-0000001163523297
+RET_CODE_PACKAGE_COMPILING = 204144727
+RET_CODE_SUBMIT_QUERY_FAILED = 204144660
```
JohanLorenzo: Genuine question: Should we make a dictionary of [all the know error codes](https://developer.huawei.com/consumer/en/doc/AppGallery-connect-References/agcapi-publishingapi-errorcode-0000001163523297) in order to make debugging easier if one day we hit them?

mozapkpublisher/huawei_api/__init__.py
```diff
@@ -7,15 +8,93 @@
...
+
+# 204144660 is reused for permanent failures (the AAB compile errors 80210099-80210116),
+# so only the message tells the two apart. Match the wording of a package still in
+# progress, never a bare "parse"/"解析": error 80210105 is the permanent "Failed to parse
+# the AAB package using the bundletool", and retrying that would bury it for the whole
+# timeout window.
+PACKAGE_PROCESSING_MESSAGES = ("is being processed", "it may take", "parsing", "解析中")
```
JohanLorenzo: Question: Where does this list come from? It feels brittle to search for these kinda generic sequence of words.

Nit: Let's add a code comment to explain what `解析中` means. Per Google Translate, it's "Analysis in progress".

mozapkpublisher/huawei_api/__init__.py
```diff
@@ -7,15 +8,93 @@
...
+    on the code alone would hide a real failure for the whole timeout window.
+    """
+    code = ret.get("code")
+    if code == RET_CODE_PACKAGE_COMPILING:
+        return True
+    if code != RET_CODE_SUBMIT_QUERY_FAILED:
+        return False
+
+    message = (ret.get("msg") or "").lower()
+    return any(fragment in message for fragment in PACKAGE_PROCESSING_MESSAGES)
```
JohanLorenzo: Code smell: Looking for this small subset of errors feels weak. Looking for generic sequences of works feels even weaker. Can we leverage the [result codes](https://developer.huawei.com/consumer/en/doc/AppGallery-connect-References/agcapi-publishingapi-errorcode-0000001163523297) more so we don't depend on string parsing? If not,  let's document why in a code comment.

mozapkpublisher/huawei_api/__init__.py
```diff
@@ -7,15 +8,93 @@
...
+
+    https://developer.huawei.com/consumer/en/doc/AppGallery-connect-References/agcapi-app-submit-0000001158245061
+    """
+    if not 0 < rollout_rate <= 100:
+        raise HuaweiUpdateException(
+            "Rollout percentage must be in (0, 100]. Value given: {}".format(rollout_rate)
+        )
+
+    start_time = start_time or datetime.now(timezone.utc)
+    percent = "{:.2f}".format(rollout_rate)
```
JohanLorenzo, line 94: Surprise: Wow, Huawei decided to send strings to encode percentage 😮 


> | Parameter | Mandatory | Type | Description |
> |------------------------|-----------|------------------------------|--|
> | phasedReleasePercent | String(10) | Phased release percentage. |  For an Android app, the value ranges from 0.00 to 100.00, accurate to two decimal places. Do not include a percent sign (%) in the value. [...] | 

I would have chosen an int since they don't go beyond 2 decimals. But anyway, there's nothing we can do.

Source https://developer.huawei.com/consumer/en/doc/AppGallery-connect-References/agcapi-app-submit-0000001158245061#section11431111271820

mozapkpublisher/huawei_api/__init__.py
```diff
@@ -61,44 +141,103 @@ async def upload_apks(self, package_name, apks, rollout_rate, submit=False):
...
+        was just bound to it.
+
+        Packages are parsed asynchronously and `app-submit` fails while that is in
+        flight, so the first attempt goes out immediately -- a small package is often
+        ready straight away -- and is then retried every `SUBMIT_RETRY_DELAY` seconds
+        until `SUBMIT_RETRY_TIMEOUT`. Any other failure is raised on the first attempt.
+
+        https://developer.huawei.com/consumer/en/doc/AppGallery-connect-References/agcapi-app-submit-0000001158245061
+        """
+        deadline = time.monotonic() + SUBMIT_RETRY_TIMEOUT
```
JohanLorenzo, line 182: TIL: `time.monotonic()`

mozapkpublisher/huawei_api/__init__.py
```diff
@@ -61,44 +141,103 @@ async def upload_apks(self, package_name, apks, rollout_rate, submit=False):
...
 
-        `fileDestUrl` is the storage reference the upload step returns for the
-        uploaded binary. It is passed to `update_app_file_info` to bind the binary
-        to the app release.
+        That URL is the storage reference the upload step returns for the uploaded
+        binary. It is passed to `update_app_file_info` to bind the binary to the app
+        release.
+
+        The upload response misspells the key as `fileDestUlr`, while the
+        `app-file-info` request that consumes it spells it `fileDestUrl`. Both
```
JohanLorenzo, line 219: Hahahaha! That's a nice piece of documentation in order to avoid confusion next time 👍

mozapkpublisher/huawei_api/__init__.py
```diff
@@ -158,6 +297,11 @@ async def get_upload_url(self, app_id: str, suffix: str = "apk", release_type: i
         Get an upload URL for a binary. When `sha256` (the hex digest of the file) is
         provided, the store verifies the integrity of the uploaded package against it.
 
+        Note the doc below now describes `/upload-url/for-obs`, which is a different
+        endpoint rather than a rename: it rejects this parameter set with
+        "Parameter is [fileName]. Parameter is required." The bare `/upload-url` route
+        used here is still live and was verified against the production API.
+
```
JohanLorenzo: Nit: Why don't we just fix the link?

huawei_auth_smoke.py
```diff
@@ -1,35 +0,0 @@
-#!/usr/bin/env python3
```
JohanLorenzo: Praise: Thanks for deleting this leftover file.

mozapkpublisher/huawei_api/__init__.py
```diff
@@ -129,24 +129,35 @@ async def upload_apks(self, package_name, apks, rollout_rate, submit=False):
...
-                await self.submit_app(app_id, RELEASE_TYPE_FULL_ROLLOUT)
-            else:
-                await self.submit_app(
-                    app_id, RELEASE_TYPE_PHASED_ROLLOUT, build_phased_release(rollout_rate)
-                )
+        if not submit:
+            logger.warning(
+                'Binaries were uploaded but NOT submitted for release. Pass `--submit` to release them.'
+            )
+            return
```
JohanLorenzo, line 154: I guess we want this for testing,right?

mozapkpublisher/huawei_api/__init__.py
```diff
@@ -129,24 +129,35 @@ async def upload_apks(self, package_name, apks, rollout_rate, submit=False):
...
+
+        if rollout_rate is None:
+            await self.submit_app(app_id, RELEASE_TYPE_FULL_ROLLOUT)
+            logger.info('Submitted app %s for full release', app_id)
```
JohanLorenzo, line 159: Nit: Let's also tell whoever reads the logs that we're going for a phased/full release before it happens. That enables them to cancel the operation if they made the wrong decision.

JohanLorenzo (approved):
This is great! Thank you very much for taking this on @hneiva!

I see this was tested live yesterday. I trust the existing code is working as intended. I found a few nits and code smell that worries me. I'm fine merging this code smell as is because I don't believe it's going to hurt us in the immediate future. However, I recommend we follow up on this, especially after pushapk publishes to the Huawei store.

mozapkpublisher/huawei_api/__init__.py
hneiva: Yes. Might also be useful for staging releases.

mozapkpublisher/huawei_api/__init__.py
hneiva: Yeah, maybe a file dedicated to those error codes would be a better solution.

mozapkpublisher/huawei_api/__init__.py
hneiva: `isoformat()` emits the offset as `+00:00`. Huawei documents `yyyy-MM-ddTHH:mm:ssZZ` with the example `2015-01-01T01:01:01+0800` without the colon in the ZZ. So `isoformat()` doesn't match the documented shape. I'll add a comment.

mozapkpublisher/huawei_api/__init__.py
hneiva: The API is undocumented 😅 

I'll update the doc to reflect it better.

mozapkpublisher/huawei_api/__init__.py
hneiva: It's a last-case scenario after all other checks have failed. I'll improve the docs.

PR #335, opened 2026-09-03
fix(workflow): Pypi publish step conflicting with build step

hneiva:
- Note: Changes the renovate config to track versions via github-tags -> semver instead of just trackign the release/v1.14 branch

The build step resolved `hatchling==1.32.0`, which writes `Metadata-Version: 2.5` (PEP 794) into the wheel.
`pypa/gh-action-pypi-publish` v1.14.0 ships `twine` v6.1, which only supports metadata versions up to 2.4, so when twine runs checks, it rejects the wheel it can't parse.
The new v1.14.2 ships with `twine` v7.0


jcristau (dismissed):
Can you explain in the commit message what this fixes and how?

ahal (approved):

